PR #60, opened 2020-12-03
Use /dev/urandom instead of /dev/random

cristiklein:
While trying out EKS-D on Ubuntu 20.04.1 LTS, the script
`./development/kops.sh` got stuck doing `cat /dev/random`.
This happened likely due to entropy exhaustion. EKS-D
needs a source of randomless to generate a kops state S3 bucket
name, hence using `urandom` is acceptable.

*Issue #, if available:* N/A

*Description of changes:*

After this change, `kops.sh` will use `/dev/urandom` instead of `/dev/random` for generating a random name for the kops S3 bucket name. This will likely improve user experience, since `/dev/random` may get stuck when system entropy is low.


micahhausler (approved):
/lgtm
/approve

Thank you for the contribution!

PR #63, opened 2020-12-04
change kops script to use template

rothgar:
*Description of changes:*

This change updates the development/kops.sh script to use a template instead of requiring manual editing of the cluster file.

This change uses a [kops cluster template](https://github.com/kubernetes/kops/blob/master/docs/operations/cluster_template.md) and generates a values.yaml file to merge config into a `$CLUSTER.yaml` file.

This also will allow us to separate out more variables from the config to allow users to deploy to different regions and change other config.

You should be able to clone this branch and run `./development/kops.sh` and be prompted for all of the necessary information.

### DIFF
--- a/development/eks-d.tpl
+++ b/development/eks-d.tpl
@@ -15,13 +15,13 @@ spec:
   etcdClusters:
   - cpuRequest: 200m
     etcdMembers:
-    - instanceGroup: master-us-west-2a
+    - instanceGroup: control-plane-{{.awsRegion}}a
       name: a
     memoryRequest: 100Mi
     name: main
   - cpuRequest: 100m
     etcdMembers:
-    - instanceGroup: master-us-west-2a
+    - instanceGroup: control-plane-{{.awsRegion}}a
       name: a
     memoryRequest: 100Mi
     name: events
@@ -42,17 +42,17 @@ spec:
   - 0.0.0.0/0
   subnets:
   - cidr: 172.20.32.0/19
-    name: us-west-2a
+    name: {{.awsRegion}}a
     type: Public
-    zone: us-west-2a
+    zone: {{.awsRegion}}a
   - cidr: 172.20.64.0/19
-    name: us-west-2b
+    name: {{.awsRegion}}b
     type: Public
-    zone: us-west-2b
+    zone: {{.awsRegion}}b
   - cidr: 172.20.96.0/19
-    name: us-west-2c
+    name: {{.awsRegion}}c
     type: Public
-    zone: us-west-2c
+    zone: {{.awsRegion}}c
   topology:
     dns:
       type: Public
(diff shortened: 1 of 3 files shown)


development/eks-d.tpl
```diff
@@ -0,0 +1,143 @@
...
+  authorization:
+    rbac: {}
+  channel: stable
+  cloudProvider: aws
+  configBase: {{ .configBase }}
+  containerRuntime: docker
+  etcdClusters:
+  - cpuRequest: 200m
+    etcdMembers:
+    - instanceGroup: master-us-west-2a
```
micahhausler, line 18: Some of the 'master' terminology is hard-coded (ex: `masterPublicName` below), but for variable names, lets use 'leader' (unless you think of something better)

And can you template out the region everywhere? 
```suggestion
    - instanceGroup: leader-{{{{.awsRegion}}}}a
```

development/eks-d.tpl
```diff
@@ -0,0 +1,143 @@
...
+apiVersion: kops.k8s.io/v1alpha2
+kind: InstanceGroup
+metadata:
+  creationTimestamp: null
+  labels:
+    kops.k8s.io/cluster: {{.clusterName}}
+  name: nodes
+spec:
+  image: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20201026
+  machineType: m5.xlarge
```
micahhausler, line 134:
```suggestion
  machineType: t3.medium
```

development/eks-d.tpl
```diff
@@ -0,0 +1,143 @@
...
+apiVersion: kops.k8s.io/v1alpha2
+kind: InstanceGroup
+metadata:
+  creationTimestamp: null
+  labels:
+    kops.k8s.io/cluster: {{.clusterName}}
+  name: master-us-west-2a
+spec:
+  image: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20201026
+  machineType: m5.large
```
micahhausler, line 114: Lets reduce the instance type, no sense in making people pay for something this expensive in a demo (I had originally set it to this for testing)
```suggestion
  machineType: t3.medium
```

micahhausler (changes requested):
Thank you for making my hack more usable!

development/eks-d.tpl
olemarkus: `control-plane` is the term we'll migrate towards in kOps, if that helps.

rothgar: I agree there were a lot of variables I wanted to pull out of the template and update names but I also didn't want to change too much in this PR to keep functionality the same and later go through and add some more options.

development/eks-d.tpl
```diff
@@ -0,0 +1,143 @@
...
+    kops.k8s.io/cluster: {{.clusterName}}
+  name: master-us-west-2a
+spec:
+  image: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20201026
+  machineType: t3.medium
+  maxSize: 1
+  minSize: 1
+  nodeLabels:
+    kops.k8s.io/instancegroup: master-us-west-2a
+  role: Master
```
rothgar, line 119: I'm assuming this is required by kops. Anyone know?

olemarkus: Yes. This one still has to be set to `Master` for control plane nodes.

micahhausler: this is littered [all over the K8s code base](https://github.com/kubernetes/kubernetes/search?q=node-role.kubernetes.io%2Fmaster), and [kops hard codes the values](https://godoc.org/k8s.io/kops/pkg/apis/kops#InstanceGroupRole)

micahhausler (approved):
/approve
/lgtm

PR #66, opened 2020-12-07
Create cluster in AWS_DEFAULT_REGION

cristiklein:
*Issue #, if available:* NA

*Description of changes:*

The kops S3 state bucket is create in $AWS_DEFAULT_REGION, however,
the Kubernetes cluster itself is created in `us-east-1`, which is
confusing.

This patch ensures that the kops configuration file creates the cluster
in the same region as the kops bucket.

### DIFF
--- a/development/kops.sh
+++ b/development/kops.sh
@@ -19,6 +19,9 @@ if [ -z "$CLUSTER_NAME" ]; then
     read -r -p "What is the name of your Cluster? " CLUSTER_NAME
 fi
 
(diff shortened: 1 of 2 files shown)


olemarkus:
Using `kops create cluster` for much beyond testing kops isn't really recommended. I would merge #63 first and then make AWS region a template variable.

cristiklein:
This was fixed more elegantly by #63 . I'm closing this.

PR #68, opened 2020-12-07
Add clarification to fix aws-iam-authenticator

rothgar:
*Description of changes:*
Just making the kops script usable with aws-iam-authenticator after the cluster has been created.

kops won't report the cluster is healthy until all pods are running so it causes the verification to wait a full 10 minutes but the pods won't schedule until the config map exists.

Still researching if we can automate this further.


micahhausler (approved):
/lgtm
/approve

PR #76, opened 2020-12-10
Move and rename the create cluster script

TerryHowe:
Simple PR to rename the create script. The docs are wrong, but once it is moved, I have some refactoring on the script so it can be automated easier.


vignesh-goutham (approved):
/lgtm
/approve

PR #80, opened 2020-12-14
Create second image tag based on date/time created

abhay-krishna:
Tag project images with a second tag while building, so that we can identify when we built them.


projects/coredns/coredns/Makefile
```diff
@@ -31,7 +31,8 @@ IMAGE_NAME?=coredns/coredns
 # This tag is overwritten in the prow job to point to the upstream git tag and this repo's commit hash
 IMAGE_TAG?=${GIT_TAG}-eks-${RELEASE_BRANCH}-${RELEASE}
 IMAGE?=$(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_TAG)
-
+DATE_EPOCH_TAG=$(shell date "+%F-%s")
```
TerryHowe, line 34: What about something like `git show --no-patch --no-notes --pretty='%ct'  $SHA1` so we get the same date for all the containers. Now that I say that though, maybe just tagging them all with the $SHA1 would be easier and more easy to correlate.

TerryHowe:
I'm going to kill this for now, feel free to revive

PR #83, opened 2020-12-14
Fix occasional problem with delete cluster

TerryHowe:
Parsing of the get cluster is failing sometimes, use this other approach seems to work better.


micahhausler (approved):
/approve

PR #89, opened 2020-12-17
Add make target to assist in updating kubernetes versions

jaxesn:
When updating kubernetes (and other projects) we want to update the expected checksums so we can validate them during the pre/post submit builds.  When generating these checksums we need to run on a linux box vs mac.  This PR adds a helper dockerfile/make target to support this workflow.  

I am open to discussion on if this a workflow we want to support or not.  

Notes
- since the kubernetes scripts pull in build/lib/common.sh the dockerfile + make target need to exist at the root so those files will be included in the docker context.
- to make copying the files out of the build image, using the buildkit backend for docker.  I'm not totally sure if this is an issue or concern or not.  If this is an issue, there is a workaround that involves spinning up a temporary container based on the image and pulling the files out that way.
- the second commit is an update to the two generated files based on running the new make target.  It looks like the existing checksums file is out of date since its pointing at old paths
- we may want to extend this process to support the other projects as well?


projects/kubernetes/kubernetes/build/create_binaries.sh
```diff
@@ -26,9 +26,9 @@ source "${MAKE_ROOT}/build/lib/init.sh"
 source "${MAKE_ROOT}/../../../build/lib/common.sh"
 
 GIT_TAG="$(cat ${MAKE_ROOT}/${RELEASE_BRANCH}/GIT_TAG)"
+RELEASE_FILE="${MAKE_ROOT}/${RELEASE_BRANCH}/RELEASE"
 PATCH_DIR=${MAKE_ROOT}/${RELEASE_BRANCH}/patches
-export KUBE_GIT_VERSION_FILE="${MAKE_ROOT}/${RELEASE_BRANCH}/KUBE_GIT_VERSION_FILE"
-
+export KUBE_GIT_VERSION=$(build::version::kube_git_version $GIT_TAG $RELEASE_FILE $RELEASE_BRANCH)
```
jaxesn, line 31: The only key that should matter during the build time process is the git_version since its reflected in the final binary.  The make the process of creating a new release easier, avoid requiring this file to exist and let the build process create it later.

projects/kubernetes/kubernetes/build/lib/version.sh
```diff
@@ -21,9 +28,10 @@ function build::version::create_env_file() {
     local -r repository="$4"
     local -r release_branch="$5"
 
-    git -C $repository checkout $tag
-    local -r source_date_epoch=$(git -C $repository show -s --format=format:%ct HEAD)
-
+    local -r source_date_epoch=$(git -C $repository show -s --format=format:%ct $tag^{commit})
```
jaxesn, line 31: This checkout confused me for a while because I was trying to run the unit tests after running this during the build and they were passing, when they should fail due to a change in one of our patches.  Changed this up to avoid changing the branch so that if we ever do need to do anything with the kubernetes source its actually at the point that we built from.

projects/kubernetes/kubernetes/build/update_checksums.sh
```diff
@@ -36,7 +36,7 @@ fi
 
 rm ${MAKE_ROOT}/${RELEASE_BRANCH}/checksums || true
 # TODO: come up with a beter filter than 'kube*'
-for file in $(find ${BIN_DIR} -name 'kube*' -type file ); do
+for file in $(find ${BIN_DIR} -name 'kube*' -type f ); do
```
jaxesn, line 39: syntax change for linux, works on mac as well

TerryHowe:
Where are we at with this change? Somehow I missed this until now.

jaxesn:
I am happy with this and open to feedback if you don't mind reviewing.

TerryHowe (approved):

PR #92, opened 2020-12-25
Run all Conformance Tests

TerryHowe:
Add script to run all conforance tests and fix some issues with kOps scripts.


TerryHowe:
I'd like to verify the Prow job is running and has the right accounts before we merge this.

TerryHowe:
Closes #96

development/kops/create_cluster.sh
```diff
@@ -18,4 +18,12 @@ set -eo pipefail
 BASEDIR=$(dirname "$0")
 source ${BASEDIR}/set_k8s_versions.sh
 
-kops update cluster $KOPS_CLUSTER_NAME --yes
+kops update cluster --yes
```
jaxesn, line 21: is the name not required here?

development/kops/run_sonobuoy.sh
```diff
@@ -26,7 +26,15 @@ if [[ "${KOPS_STATE_STORE}" != s3://* ]]
 then
   export KOPS_STATE_STORE="s3://${KOPS_STATE_STORE}"
 fi
-export KOPS_CLUSTER_NAME=$(kops get cluster --state "${KOPS_STATE_STORE}" | tail -n +2 | cut -f1 -d '  ' 2>/dev/null)
+if [ -z "${KOPS_CLUSTER_NAME}" ]
+then
+  export KOPS_CLUSTER_NAME=$(kops get cluster --state "${KOPS_STATE_STORE}" | tail -n +2 | cut -f1 -d '	' 2>/dev/null)
```
jaxesn, line 31: were the extra spaces after the `-d` intentional here?

development/kops/run_sonobuoy.sh
```diff
@@ -38,10 +46,12 @@ rm -f LICENSE
 chmod 755 sonobuoy
 
 echo "Testing cluster $KOPS_STATE_STORE"
-export KOPS_FEATURE_FLAGS=SpecOverrideFlag
-kops set cluster "${KOPS_CLUSTER_NAME}" cluster.spec.nodePortAccess=0.0.0.0/0
 ./sonobuoy run --mode=certified-conformance --wait --kube-conformance-image k8s.gcr.io/conformance:v1.18.9
```
jaxesn, line 49: instead of hardcoding v1.18.9 here, could we get that from an environment var? maybe [GIT_TAG](https://github.com/aws/eks-distro/blob/main/projects/kubernetes/kubernetes/Makefile#L5)

jaxesn (changes requested):
Seems pretty reasonable to me, some minor questions.

development/kops/create_cluster.sh
TerryHowe: It gets it from the environment

development/kops/run_sonobuoy.sh
TerryHowe: That is actually a tab I think, cut only allows one character for -d as I recall

development/kops/run_sonobuoy.sh
TerryHowe: We have another ticket for pulling these artifacts from s3 or ECR

development/kops/create_cluster.sh
jaxesn: 👍

TerryHowe:
This is good to go now. I had a merge conflict which was one change. I also realized I should make some changes to the create cluster when I was preparing for the demo.

development/kops/cluster_wait.sh
```diff
@@ -13,8 +13,26 @@
...
+        exit 1
+    fi
+    echo 'Waiting for cluster to come up...'
+done
+
+set -x
+kops validate cluster --wait 3m
+
+# This may not be required
+# kubectl delete pod -n kube-system -l k8s-app=aws-iam-authenticator
```
vivek-koppuru, line 38: So is this actually required? If we are leaving it as a reference, let's explain that a bit more in the comment above

development/kops/create_configuration.sh
```diff
@@ -81,26 +82,26 @@ data:
...
 export SSH_KEY_PATH=${SSH_KEY_PATH:-$HOME/.ssh/id_rsa.pub}
 kops create secret --name $KOPS_CLUSTER_NAME sshpublickey admin -i ${SSH_KEY_PATH}
 
 echo
-echo "# Set these values"
-echo "export AWS_DEFAULT_REGION=$AWS_DEFAULT_REGION"
-echo "export KOPS_CLUSTER_NAME=$KOPS_CLUSTER_NAME"
-echo "export KOPS_STATE_STORE=$KOPS_STATE_STORE"
+echo "# Creating ./${KOPS_CLUSTER_NAME}/env.sh"
+echo "export AWS_DEFAULT_REGION=$AWS_DEFAULT_REGION" | tee ./${KOPS_CLUSTER_NAME}/env.sh
```
vivek-koppuru, line 105: Should we be checking for AWS_REGION here before setting AWS_DEFAULT_REGION, or keeping it inline with the prow.sh in the other PR?

development/kops/cluster_wait.sh
TerryHowe: It doesn't seem to be needed. Conformance tests pass without a restart. Old comment I guess.

development/kops/create_configuration.sh
TerryHowe: Okay, I updated to it now supports old and new region env vars. Thanks!

development/kops/cluster_wait.sh
TerryHowe: Removed that comment.

TerryHowe:
Closes: https://github.com/aws/eks-distro/issues/115

PR #94, opened 2020-12-28
Add prow.sh script

TerryHowe:
This script will set up the credentials for Prow.

### DIFF
--- a/development/kops/prow.sh
+++ b/development/kops/prow.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+# Copyright 2020 Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+if [ "$AWS_ROLE_ARN" == "" ]; then
+    echo "Empty AWS_ROLE_ARN, this script must be run in a postsubmit pod with IAM Roles for Service Accounts"
+    exit 1
+fi
+
+BASEDIR=$(dirname "$0")
+
+cat << EOF > ${BASEDIR}/config
+[default]
+output = json
+region = us-west-2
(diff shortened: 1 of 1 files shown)


development/kops/prow.sh
```diff
@@ -0,0 +1,39 @@
...
+fi
+
+BASEDIR=$(dirname "$0")
+
+cat << EOF > ${BASEDIR}/config
+[default]
+output = json
+region = us-west-2
+role_arn=$AWS_ROLE_ARN
+web_identity_token_file=/var/run/secrets/eks.amazonaws.com/serviceaccount/token
```
TerryHowe, line 28: I don't know the name of the secret. That probably isn't correct.

development/kops/prow.sh
```diff
@@ -0,0 +1,45 @@
...
+cat << EOF > ${BASEDIR}/config
+[default]
+output=json
+region=us-west-2
+role_arn=$PROW_ROLE_ARN
+web_identity_token_file=/var/run/secrets/eks.amazonaws.com/serviceaccount/token
+
+[profile conformance-test]
+role_arn=$TEST_ROLE_ARN
+region=us-west-2
```
micahhausler, line 37: Can this be parameterized?

micahhausler (changes requested):

development/kops/prow.sh
TerryHowe: The region?

micahhausler: Yes. If kops is running in `us-east-1` wouldn't you want to set that here?

TerryHowe: Latest change allows overriding default of us-west-2 with AWS_REGION or AWS_DEFAULT_REGION

vivek-koppuru: So the region would be specified in the prowjob itself, and the customer would do that by setting the environment variable before running the script? I think it's okay to have us-west-2 as default as this is specifically for our prow job, but the other options is to actually pass in the region as a string to this script.

TerryHowe: Done

development/kops/prow.sh
```diff
@@ -0,0 +1,39 @@
...
+
+cat << EOF > ${BASEDIR}/config
+[default]
+output = json
+region = us-west-2
+role_arn=$AWS_ROLE_ARN
+web_identity_token_file=/var/run/secrets/eks.amazonaws.com/serviceaccount/token
+
+[profile conformance-test]
+role_arn = arn:aws:iam::379412251201:role/DocsDeploymentRole
```
bnrjee: Should we use ArtifactsDeploymentRole instead?

abhay-krishna (approved):
/lgtm
/approve

PR #103, opened 2021-01-04
Minor make file fixes

TerryHowe:
Build problems I found fixed.

### DIFF
--- a/projects/kubernetes/kubernetes/Makefile
+++ b/projects/kubernetes/kubernetes/Makefile
@@ -8,7 +8,7 @@ AWS_ACCOUNT_ID?=$(shell aws sts get-caller-identity --query Account --output tex
 AWS_REGION?=us-west-2
 
 RELEASE_BRANCH?="1-18"
-RELEASE?="1"
+RELEASE?=1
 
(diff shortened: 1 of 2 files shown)


projects/kubernetes/kubernetes/Makefile
```diff
@@ -50,7 +50,6 @@ pause:
 	mkdir -p $(PAUSE_DST_DIR)
 	# Older versions have the pause.c file stored under kubernetes/build/pause/
 	test -f $(PAUSE_SRC_DIR)/linux/pause.c && cp $(PAUSE_SRC_DIR)/linux/pause.c $(PAUSE_DST_DIR) || cp $(PAUSE_SRC_DIR)/pause.c $(PAUSE_DST_DIR)
-	rm -rf ./kubernetes
```
jaxesn, line 53: What was the issue/motivation here?

TerryHowe: That gets used in a step later on and it is missing.  It is something that should probably be done, but not here.

TerryHowe: I'm going to move this change to another PR. I assume that was put there because of the space issue for tests and I'll come up with some other fix for that.

vivek-koppuru (approved):
/lgtm

PR #105, opened 2021-01-05
Set GOPROXY

bnrjee:
*Issue #, if available:*

*Description of changes:*


TerryHowe:
Was there a problem where Prow couldn't just set and export GO_PROXY for the job? All this wouldn't be necessary if that was done.

bnrjee:
> Was there a problem where Prow couldn't just set and export GO_PROXY for the job? All this wouldn't be necessary if that was done.

We just need to set GOPROXY to direct for 2 of the csi targets. Prow preset will set the GOPROXY value to the athens service dns. For these 2 targets, we need to use direct, hence this change.

projects/kubernetes-csi/external-provisioner/Makefile
```diff
@@ -30,6 +30,8 @@ IMAGE_REPO?=$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com
 IMAGE_NAME?=kubernetes-csi/external-provisioner
 IMAGE_TAG?=$(GIT_TAG)-eks-${RELEASE_BRANCH}-${RELEASE}
 IMAGE?=$(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_TAG)
+GOPROXY_DNS?=direct
+export GOPROXY=$(GOPROXY_DNS)
```
micahhausler, line 34: Why do you need to variables? Can you not just do
```Makefile
export GOPROXY?=direct
```

Same for below

bnrjee: I tested your the suggested changes. It does not set the proxy without the variable.

micahhausler (approved):
/lgtm
/approve

Issue #99, opened 2021-01-02
ARM64 architecture support for RPi testing


paavan98pm:
**What would you like to be added**:
Add ARM64 support.

**Why is this needed**:
Since Microk8s already supports this for Raspberry Pi installs and testing, this will help a number of customers, given the EKS Anywhere announcement. https://microk8s.io/docs/install-alternatives#heading--arm

TerryHowe:
I haven't tried arm64, but it should work https://gallery.ecr.aws/eks-distro/kubernetes/kube-apiserver for example notice it shows arm64

jaxesn:
As @TerryHowe mentioned, arm64 builds are available for the EKS-D binaries/container images. @ktsakalozos might be able to confirm if the eks snap supports arm64 or not?

ktsakalozos:
At this point the EKS snap is for AMD64 only, ARM support should come soon.

TerryHowe:
The images definitely work on 64 bit Arm, Buster:
```root@rpi-159:~# uname -a
Linux rpi-159 5.4.79-v8+ #1373 SMP PREEMPT Mon Nov 23 13:32:41 GMT 2020 aarch64 GNU/Linux
root@rpi-159:~# cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```
I have some etcd config problems, so not a running cluster at this point.

ktsakalozos:
@paavan98pm to get the EKS snap on ARM64 currently you will to build it on an ARM^$ machine with:
```
sudo snap install snapcraft --classic
git clone https://github.com/canonical/eks-snap.git
cd eks-snap
snapcraft
```
From a quick test I did, all seemed to work apart from the storage integration with AWS.

paavan98pm:
@ktsakalozos did you need KVM-support instance for this? It also asked for multipass. thx

paavan98pm:
Both, R-Pi and c6g.metal returned the error below when running `snapcraft`:

```
launch failed: Internal error: qemu-system-x86_64 failed getting vmstate (Process returned exit code: 1) with output:
[...]

PR #106, opened 2021-01-05
etcd arm64 support fix

TerryHowe:
The etcd container fails immediately on arm64 unless the ETCD_UNSUPPORTED_ARCH environment variable is set. The work around is do something like:

    docker run -e ETCD_UNSUPPORTED_ARCH=arm64  public.ecr.aws/j0a1m4z9/etcd-io/etcd:v3.4.14-eks-1-18-1

This work around does not work when you are using kubeadm because kubeadm doesn't give you a way to set an environment variable for etcd. With this change, etcd just logs a warning that you are using an experimental version of etcd.

Related:
https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/supported-platform.md

### DIFF
--- a/projects/etcd-io/etcd/docker/linux/Dockerfile
+++ b/projects/etcd-io/etcd/docker/linux/Dockerfile
@@ -13,10 +13,17 @@
 # limitations under the License.
 
 ARG BASE_IMAGE
-FROM $BASE_IMAGE
+FROM $BASE_IMAGE AS base
 
 ARG TARGETARCH
 ARG TARGETOS
+
+FROM base AS branch-version-amd64
+
+FROM base AS branch-version-arm64
+ENV ETCD_UNSUPPORTED_ARCH=arm64
+
+FROM branch-version-${TARGETARCH} AS final
 
 COPY LICENSES /LICENSES
 COPY ATTRIBUTION.txt /ATTRIBUTION.txt


projects/etcd-io/etcd/docker/linux/Dockerfile
```diff
@@ -13,11 +13,18 @@
...
 
 ARG TARGETARCH
 ARG TARGETOS
 
+FROM base AS branch-version-amd64
+
+FROM base AS branch-version-arm64
+ENV ETCD_UNSUPPORTED_ARCH=arm64
+
+FROM branch-version-${TARGETARCH} AS final
```
jaxesn, line 26: Is `final` special here?  Im just now sure how the process running the build knows to that tag that as the actual image?

jaxesn (commented):
Seems like a smart approach!  Just a question on multi-stage build process

projects/etcd-io/etcd/docker/linux/Dockerfile
TerryHowe: I don't think final is a key word or anything. I look at that as more of a comment to let you know that is the final stage of a multistage build.

As I understand it, arm64 and amd64 both get tagged the same and when you pull, it attempts to pull the best match for your architecture. When I'm on my mac I run a `docker pull` of one of those images and I get an amd64 version. On the Pi I the same command gives me the arm64 version.

TerryHowe:
This may help https://github.com/aws/eks-distro/issues/99

projects/etcd-io/etcd/docker/linux/Dockerfile
jaxesn: ok that makes sense

micahhausler (approved):
/lgtm
/approve

PR #108, opened 2021-01-06
Tag pause container 3.2

TerryHowe:
This change fixes kubeadm which has 3.2 hard coded as the version of pause to deploy.

https://github.com/kubernetes/kubernetes/blob/release-1.18/cmd/kubeadm/app/constants/constants_unix.go#L26

### DIFF
--- a/projects/kubernetes/kubernetes/Makefile
+++ b/projects/kubernetes/kubernetes/Makefile
@@ -86,11 +86,13 @@ docker: binaries pause
 		--build-arg BASE_IMAGE=$(GO_RUNNER_IMAGE) \
 		--build-arg VERSION=$(IMAGE_TAG) \
 		-t $(PAUSE_IMAGE) \
+		-t $(IMAGE_REPO)/$(IMAGE_REPO_PREFIX)/$(PAUSE_IMAGE_NAME):3.2 \
 		-f ./docker/pause/Dockerfile ./_output/pause/
 
 .PHONY: docker-push
 docker-push:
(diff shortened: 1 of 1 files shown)


TerryHowe:
This may also assist with https://github.com/aws/eks-distro/issues/99

projects/kubernetes/kubernetes/Makefile
```diff
@@ -86,11 +86,13 @@ docker: binaries pause
 		--build-arg BASE_IMAGE=$(GO_RUNNER_IMAGE) \
 		--build-arg VERSION=$(IMAGE_TAG) \
 		-t $(PAUSE_IMAGE) \
+		-t $(IMAGE_REPO)/$(IMAGE_REPO_PREFIX)/$(PAUSE_IMAGE_NAME):3.2 \
```
jaxesn, line 89: Any idea how often the version of pause upstream changes?  Is 3.2 something that we need to be looking out to change when updating kub versions?

TerryHowe: 1.20 k8s still uses 3.2 pause. We won't be changing it often.

TerryHowe: I'm going to look at making kubeadm configurable for pause version, but that won't be available (if I get something through) until 1.21.

jaxesn: sgtm

jaxesn (approved):

vivek-koppuru (approved):
/lgtm
/approve

PR #109, opened 2021-01-07
Add docker push target for kubernetes build

TerryHowe:
project/kubernetes/kubernetes did not have a docker-push target

### DIFF
--- a/projects/kubernetes/kubernetes/Makefile
+++ b/projects/kubernetes/kubernetes/Makefile
@@ -90,6 +90,7 @@ docker: binaries pause
 
 .PHONY: docker-push
 docker-push:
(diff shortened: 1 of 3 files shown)


jaxesn (approved):
Seems reasonable to me.  Just wondering if this introduces a too easy way to accidentally push up image to prod?  I'm thinking not really since we'd have to have those creds locally?

TerryHowe:
> Seems reasonable to me. Just wondering if this introduces a too easy way to accidentally push up image to prod? I'm thinking not really since we'd have to have those creds locally?

Pushing to prod would require aws cli access (which would trigger a ticket) and docker login/config. If it happened, we'd know about it.

PR #114, opened 2021-01-12
Use specific go version binaries in each project

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

Adding ability to specify the version of go we want for each project. This PR will only use the different version of go once the following PR is merged: https://github.com/aws/eks-distro-prow-jobs/pull/69

For most projects, there is a variable we set in the create_binaries script that will look for the appropriate version of go. Note that we don't have to specify the 3rd number in the list of numbers as we set that in the common script, allowing us to not have to change the version for every single project if we are just doing a go upgrade.

/hold

### DIFF
--- a/build/lib/common.sh
+++ b/build/lib/common.sh
@@ -76,3 +76,28 @@ function build::gather_licenses() {
      -o -name 'NOTICE' -o -name 'COPYING' -o -name 'NOTICE.txt' \
      -o -name 'LICENSE.code' -o -name 'LICENCE.md' \) | while IFS= read -r NAME; do mkdir -p "${outputdir}/$(dirname $NAME)" && cp  "$NAME" "${outputdir}/${NAME}"; done
 }
+
+function build::common::use_go_version() {
+  local -r version=$1
+  local gobinaryversion=""
+
+  if [[ $version == "1.13"* ]]; then
+    gobinaryversion="1.13"
+  fi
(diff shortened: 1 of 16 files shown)


build/lib/common.sh
```diff
@@ -76,3 +76,30 @@ function build::gather_licenses() {
      -o -name 'NOTICE' -o -name 'COPYING' -o -name 'NOTICE.txt' \
      -o -name 'LICENSE.code' -o -name 'LICENCE.md' \) | while IFS= read -r NAME; do mkdir -p "${outputdir}/$(dirname $NAME)" && cp  "$NAME" "${outputdir}/${NAME}"; done
 }
+
+function build::common::use_go_version() {
```
jaxesn, line 80: Minor nit, it would be nice if when adding golang versions to builder base, we created a symlink to like /what/ever/bin/go1.14.13 -> go1.14, that way in this repo we don't have to know the specific patch versions for each minor version.

vivek-koppuru: I get what you are saying, that's a good idea. I would be creating the symlinks on the install side so would need to create another PR for the build-tooling repo but I can do that.

jaxesn (approved):
I built 1.19 using this and it correctly used the 1.15 version.

vivek-koppuru:
Tests running fine now, can be merged.

PR #117, opened 2021-01-12
Artifacts

TerryHowe:
Determine the latest tagged artifacts to use for testing.

### DIFF
--- a/development/kops/create_values_yaml.sh
+++ b/development/kops/create_values_yaml.sh
@@ -37,18 +37,37 @@ fi
 
 export DEFAULT_REPOSITORY_URI=public.ecr.aws/eks-distro
 export REPOSITORY_URI=${REPOSITORY_URI:-${DEFAULT_REPOSITORY_URI}}
-function get_container_yaml() {
+function get_container_latest_tag() {
     REPOSITORY_NAME="${1}"
-    TAG="${2}"
-    if [ "${REPOSITORY_URI}" != "${DEFAULT_REPOSITORY_URI}" ]
+    DEFAULT_TAG="${2}"
+    if [ "${REPOSITORY_URI}" == "${DEFAULT_REPOSITORY_URI}" ]
     then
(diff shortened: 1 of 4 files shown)


development/kops/create_values_yaml.sh
```diff
@@ -0,0 +1,99 @@
...
+}
+
+echo "Creating ./${KOPS_CLUSTER_NAME}/values.yaml"
+cat << EOF > ./${KOPS_CLUSTER_NAME}/values.yaml
+kubernetesVersion: $KUBERNETES_VERSION
+clusterName: $KOPS_CLUSTER_NAME
+configBase: $KOPS_STATE_STORE/$KOPS_CLUSTER_NAME
+awsRegion: $AWS_DEFAULT_REGION
+pause:
+$(get_container_yaml kubernetes/pause v1.18.9-eks-1-18-1)
```
jaxesn, line 84: Could we make `v1.18.9-eks-1-18-1` configurable via a env var?  In the builds for 1.18 vs 1.19 we could pass this through to control which images we want to test against.  There may be a few other places in these scripts, like set_k8s_versions.sh, that we also will need to look at.

TerryHowe: The version passed to get_container_yaml is the default version. Yeh, something else needs to be done here, but this preserves the way it worked before for now. Personally, I think we should build a CRD every time and tag that with the SHA1 and use that artifact to determine what containers go here, but that is kind of out of scope of this e2e project.

jaxesn: Ok I think I see.  The way this is written its literally finding the absolute latest tag?  So if we had multiple "channels", 1.18 vs 1.19, it would just always find the latest 1.19? or the latest by date across all versions?

TerryHowe: Yes, absolute latest tag and it will work fo now and get some tests running.

I personally don't like this approach, I like the CRD approach or more intelligent tagging in ECR better. Either of those approaches will take longer though.

jaxesn: makes sense

vivek-koppuru (approved):

PR #122, opened 2021-01-14
Track eks-distro-base image in separate file

abhay-krishna:
Tracking the tag for EKS Distro base image in a separate file, so that automatic PR logic will result in lighter PR's since just one file will change and not 12+ Makefiles.


Makefile
```diff
@@ -4,7 +4,7 @@ DEVELOPMENT?=false
 AWS_ACCOUNT_ID?=$(shell aws sts get-caller-identity --query Account --output text)
 AWS_REGION?=us-west-2
 IMAGE_REPO?=$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com
-BASE_IMAGE?=$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com/eks-distro/base:2021-01-05-1609822390
+BASE_IMAGE?=$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com/eks-distro/base:$(shell cat EKS_DISTRO_BASE_TAG_FILE)
```
vivek-koppuru, line 7: Let's separate out the variable and have that overridable instead, like BASE_IMAGE_TAG or BASE_TAG like below

vivek-koppuru (changes requested):

Makefile
abhay-krishna: Sure

vivek-koppuru (approved):
/lgtm

PR #124, opened 2021-01-14
fix kubernetes patches to enable make test in builds

jaxesn:
This adds kubernetes `make test` to the presubmit builds.  Does not currently add it to the release/postsubmit build.

make test is the upstream kubernetes make test target.  It runs more than just "unit tests", some of the tests are spinning up the api server, utilizing etcd, etc.  I say this because there are Flakes 😱   Upstream does a great job tracking and tackling, see their approach [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/flaky-tests.md). [Here](http://storage.googleapis.com/k8s-metrics/flakes-latest.json) is their latest top flakes report, the `ci-kubernetes-cached-make-test` job is equivalent to what we are running here.

To avoid bogging us down with failed builds due to upstream flakes, this PR adds basic retry "logic".  Upstream already caches successfully test runs so when running `make test` over and over, it only runs the previously failed tests.  After running the tests in prow a number of times, up to 3 retries seems to be enough to get everything passing, but I think something we should watch for and potentially increase if necessary.  Funny enough, 1.19 actually seems flakier than 1.18 and the worst offenders match upstreams worst offenders.

### DIFF
--- a/build/update-kubernetes-version/Dockerfile
+++ b/build/update-kubernetes-version/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM 316434458148.dkr.ecr.us-west-2.amazonaws.com/eks-distro/builder:e35d4971088ca37db784caf525373f7342126bf3 as build
+FROM 316434458148.dkr.ecr.us-west-2.amazonaws.com/eks-distro/builder:afa151c76fdefe7248c1a75bb1a764c7e28fb334 as build
 ARG RELEASE_BRANCH 
 ARG KUBERNETES_ROOT
 
--- a/projects/kubernetes/kubernetes/1-18/patches/0010-EKS-PATCH-refine-aws-loadbalancer-worker-node-SG-rul.patch
+++ b/projects/kubernetes/kubernetes/1-18/patches/0010-EKS-PATCH-refine-aws-loadbalancer-worker-node-SG-rul.patch
@@ -117,64 +117,10 @@ diff --git a/staging/src/k8s.io/legacy-cloud-providers/aws/aws_test.go b/staging
 index b4549a80527..720de76d58e 100644
 --- a/staging/src/k8s.io/legacy-cloud-providers/aws/aws_test.go
 +++ b/staging/src/k8s.io/legacy-cloud-providers/aws/aws_test.go
-@@ -2024,3 +2024,116 @@ func TestConstructStsEndpoint(t *testing.T) {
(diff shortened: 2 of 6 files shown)


projects/kubernetes/kubernetes/1-18/patches/0010-EKS-PATCH-refine-aws-loadbalancer-worker-node-SG-rul.patch
```diff
@@ -117,65 +117,11 @@ diff --git a/staging/src/k8s.io/legacy-cloud-providers/aws/aws_test.go b/staging
 index b4549a80527..720de76d58e 100644
 --- a/staging/src/k8s.io/legacy-cloud-providers/aws/aws_test.go
 +++ b/staging/src/k8s.io/legacy-cloud-providers/aws/aws_test.go
-@@ -2024,3 +2024,116 @@ func TestConstructStsEndpoint(t *testing.T) {
+@@ -2024,3 +2024,62 @@ func TestConstructStsEndpoint(t *testing.T) {
  		require.NoError(t, err)
  	})
  }
 +
```
jaxesn, line 124: These were some fargate tests that slipped in unnoticed because we were never building the test files.

build/update-kubernetes-version/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM 316434458148.dkr.ecr.us-west-2.amazonaws.com/eks-distro/builder:e35d4971088ca37db784caf525373f7342126bf3 as build
+FROM 316434458148.dkr.ecr.us-west-2.amazonaws.com/eks-distro/builder:afa151c76fdefe7248c1a75bb1a764c7e28fb334 as build
```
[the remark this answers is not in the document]
vivek-koppuru: Do we actually leverage the tools in the build image? If not, wondering if we can take a dependency on our base image instead, or even better the base image of eks-distro?

projects/kubernetes/kubernetes/1-18/patches/0010-EKS-PATCH-refine-aws-loadbalancer-worker-node-SG-rul.patch
vivek-koppuru: Good catch 💯

vivek-koppuru (approved):
/lgtm

PR #125, opened 2021-01-15
WIP: add step to produce KIND images for EKS-D

chanwit:
*Issue #, if available:*

*Description of changes:*

This PR adds a sub-project to produce KIND-based images for EKS-D, by 
  [ ] copying binaries from the previous build steps (kubernetes/_output) into the sub-project directory
  [ ] changing the version of KIND image to match with the version-build of EKS-D.
  [ ] build the image, and push to the target repository

### DIFF
--- a/projects/kind-images/Dockerfile.amd64
+++ b/projects/kind-images/Dockerfile.amd64
@@ -1,0 +1,8 @@
+FROM kindest/node:v1.18.8
+
+ARG K8S_VERSION=v1.18.9-eks-1-18-1
+
+COPY amd64/* /usr/bin/
+
+RUN echo ${K8S_VERSION} > /kind/version
+RUN chmod +x /usr/bin/kubectl /usr/bin/kubeadm /usr/bin/kubelet
--- a/Makefile
+++ b/Makefile
@@ -84,3 +84,5 @@ makes:
 	$(call presubmit-cleanup, $(TARGET), "projects/kubernetes-csi/external-snapshotter")
 	make -C projects/kubernetes-csi/external-provisioner $(TARGET)
 	$(call presubmit-cleanup, $(TARGET), "projects/kubernetes-csi/external-provisioner")
+	make -C projects/kind-images $(TARGET)
+	$(call presubmit-cleanup, $(TARGET), "projects/kind-images")


projects/kind-images/Dockerfile.amd64
```diff
@@ -0,0 +1,8 @@
+FROM kindest/node:v1.18.8
```
micahhausler, line 1: Can you be sure to build this image from source?

chanwit: Hi Micah,
I'm and looking for the best approach we tackle this problem together.
This is a point I'd like to discuss too.

Would you like it to be built directly from the KIND base image?
or
Would you like it to go from a ready-to-use KIND node image, and just simply replace the binaries from EKS-D?

My POC was working great using the latter case, but I'm very open to any approach.

chanwit: Build a KIND node image from source means we'll need to re-compile binaries again from the whole Kubernetes source tree.
But I'd like to use the compiled binaries from the previous steps to make sure that *we actually ship exactly the same images*.

chanwit: Also according to the underlying implementation of the KIND build command,
https://github.com/kubernetes-sigs/kind/blob/master/pkg/build/nodeimage/build_impl.go#L90
  - copy required binaries to the container
  - chmod +x them
  - create the version file
  - and just commit to create the new image.

The Dockerfile in this PR does just that.

micahhausler: For EKS Distro, we don't distribute images or binaries we have not built from source. 

> Would you like it to be built directly from the KIND base image?
or
Would you like it to go from a ready-to-use KIND node image, and just simply replace the binaries from EKS-D?

We'd want to build the `kind` binary and image from source, so neither of these options quite work. Looking at the Kind docs on the [node image](https://kind.sigs.k8s.io/docs/design/node-image), I think we'd need to build the [base image](https://github.com/kubernetes-sigs/kind/blob/master/images/base/Dockerfile), and then invoke `kind build node-image` referencing our base image.

Looking at the kind source, we'll probably (independently from this change) want to have [`kind build node-image --type`](https://github.com/kubernetes-sigs/kind/blob/master/pkg/cmd/kind/build/nodeimage/nodeimage.go)   support the a new option (Maybe some YAML config file?) that satisfies the [`Builder`](https://github.com/kubernetes-sigs/kind/blob/master/pkg/build/nodeimage/internal/kube/builder.go#L26-L31) and [`Bits`](https://github.com/kubernetes-sigs/kind/blob/master/pkg/build/nodeimage/internal/kube/bits.go#L22-L31) interfaces. EKS-D builds the binary images and image tars, so there is no need for `kind build nodeimage` itself to need to build K8s from source.

micahhausler: Right! 

0. Yes! I don't think it would be a heavy lift to have a YAML config-based `Builder()` that could be specified from the command line. I think there would be advantage in adding the OCI tars we provide, as kind will [side-load the images](https://github.com/kubernetes-sigs/kind/blob/8ea030becbb6e0ef9012b0631e892f7d15281962/pkg/build/nodeimage/build_impl.go#L142-L153) and use those by default 
1. Lets go with our convention of `/projects/<github org>/<repo>`, so `projects/kubernetes-sigs/kind/`
2. The current kind base image [builds from Ubuntu](https://github.com/kubernetes-sigs/kind/blob/master/images/base/Dockerfile) and our preference would be to go with Amazon Linux 2. It doesn't look like it would be a heavy lift to move to an AL2 image and install containerd/runc from `amazon-linux-extras`. [`crictl`](https://github.com/kubernetes-sigs/kind/blob/master/images/base/Dockerfile#L96-L100) might be the only thing in that base image that neither Amazon Linux nor EKS-distro (yet) builds today.
3. Correct, although we'll have to figure out a solution for this too as EKS-Distro build jobs don't have Docker available and instead use buildkit for container images.

micahhausler: I created #130 to track this

chanwit:
I close this PR and will create a new one after we have a good solution to implement KinD-based images.

PR #126, opened 2021-01-15
Add main postsubmit job

bnrjee:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cmd/main_postsubmit.go
+++ b/cmd/main_postsubmit.go
@@ -1,0 +1,128 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"strconv"
+	"strings"
+)
+
+type Command struct {
+	releaseBranch        string
+	gitRoot              string
+	release              string
+	artifactBucket       string
+	uploadToPublicBucket bool
+	args                 []string
+}
+
+func (cmd Command) run(runMake bool, uploadArtifacts bool, targetPath string) {
+	if runMake == true {
+		// Add a specific target
+		cmd.args[1] = cmd.args[1] + targetPath
+		command := exec.Command("make", cmd.args...)
+		output, err := command.CombinedOutput()
+		if err != nil {
+			log.Fatalf("There was an error running make: %v. Make output:\n%v", err, string(output))
+		}
+		fmt.Printf("Output of the make command for %v:\n %v", targetPath, string(output))
+		// Remove the target name from the build args
+		cmd.args[1] = cmd.gitRoot + "/projects/"
+		if uploadArtifacts == true {
+			if cmd.uploadToPublicBucket == true {
+				command = exec.Command("/bin/bash", cmd.gitRoot+"/release/lib/create_final_dir.sh", cmd.releaseBranch, cmd.release, cmd.artifactBucket, targetPath)
+				output, err = command.CombinedOutput()
+				if err != nil {
+					log.Fatalf("There was an error running the create_final_dir script: %v. Output:\n%v", err, string(output))
+				}
+				fmt.Printf("Output of the create_final_dir script for %v:\n %v", targetPath, string(output))
+			}
+			if targetPath == "kubernetes/kubernetes" {
+				command = exec.Command("/bin/bash", "-c", fmt.Sprintf("mv %s/projects/%s/_output/%s/* /logs/artifacts", cmd.gitRoot, targetPath, cmd.releaseBranch))
+			} else {
+				command = exec.Command("/bin/bash", "-c", fmt.Sprintf("mv %s/projects/%s/_output/tar/* /logs/artifacts", cmd.gitRoot, targetPath))
+			}
+			output, err = command.CombinedOutput()
+			if err != nil {
+				log.Fatalf("There was an error running mv: %v. Output:\n%v", err, string(output))
+			}
+			fmt.Printf("Successfully moved artifacts to /logs/artifacts directory for %v.\n", targetPath)
+		}
+	}
+}
+
+func NewCommand() *Command {
+	cmd := new(Command)
+	cmd.releaseBranch = os.Args[2]
+	cmd.release = os.Args[3]
+	cmd.artifactBucket = os.Args[11]
+	cmd.uploadToPublicBucket, _ = strconv.ParseBool(os.Args[12])
+	gitRootOutput, err := exec.Command("git", "rev-parse", "--show-toplevel").Output()
+	if err != nil {
+		log.Fatalf("There was an error running the git command: %v", err)
+	}
+	cmd.gitRoot = strings.Fields(string(gitRootOutput))[0]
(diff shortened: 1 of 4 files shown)


cmd/main_postsubmit.go
```diff
@@ -0,0 +1,99 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+)
+
```
TerryHowe, line 10: What you've done here looks like it will work just fine, but I think a bit more time should be put into design here for a couple reasons. First I think this will probably grow and second we have a little breathing room right now and we should start some good design habits before the rush comes.

This could benefit from some object oriented design I think so we have some separation of concerns better defined. For example the main noun we are dealing with here seems to be "Command". Something like:
```
type Command struct {
   releaseBranch string
   gitRoot string
   release string
   artifactBucket string
   args []string
}

func NewCommand() *Command {
    cmd := new(Command)
    cmd.releaseBranch = os.Args[2]
    cmd.release = os.Args[3]
    cmd.artifactBucket = os.Args[11]
    gitRootOutput, err := exec.Command("git", "rev-parse", "--show-toplevel").Output()
    if err != nil {
        log.Fatalf("There was an error running the git command: %v", err)
    }
    gitRoot := strings.Fields(string(gitRootOutput))[0]
    fmt.Println(gitRoot)
    cmd.args = []string{"-C", gitRoot + "/projects/", os.Args[1],
        "RELEASE_BRANCH=" + os.Args[2],
        "RELEASE=" + os.Args[3],
        "DEVELOPMENT=" + os.Args[4],
        "AWS_REGION=" + os.Args[5],
        "AWS_ACCOUNT_ID=" + os.Args[6],
        "BASE_IMAGE=" + os.Args[7],
        "IMAGE_REPO=" + os.Args[8],
        "GO_RUNNER_IMAGE=" + os.Args[9],
        "KUBE_PROXY_BASE_IMAGE=" + os.Args[10],
        "IMAGE_TAG='$(GIT_TAG)-$(PULL_BASE_SHA)'"}
    return cmd
}
```

cmd/main_postsubmit.go
```diff
@@ -0,0 +1,99 @@
...
+func runCommand(root string, runMake bool, path string, releaseBranch string, release string, artifactBucket string, args []string) {
```
TerryHowe, line 11: Then this guy, the arguments just need to be converted to get from `cmd` and it becomes `func (cmd *Command) run(runMake bool, path string) int {`

cmd/main_postsubmit.go
```diff
@@ -0,0 +1,99 @@
...
+			iamAuthChanged = true
+			etcdChanged = true
+		}
+	}
+	buildArg := []string{"-C", gitRoot + "/projects/", os.Args[1],
+		"RELEASE_BRANCH=" + os.Args[2], "RELEASE=" + os.Args[3],
+		"DEVELOPMENT=" + os.Args[4], "AWS_REGION=" + os.Args[5],
+		"AWS_ACCOUNT_ID=" + os.Args[6], "BASE_IMAGE=" + os.Args[7],
+		"IMAGE_REPO=" + os.Args[8], "IMAGE_TAG='$(GIT_TAG)-$(PULL_BASE_SHA)'"}
+	kubeBuildArg := []string{"-C", gitRoot + "/projects/", os.Args[1],
```
TerryHowe, line 89: It seems like we could consolidate these two since passing extra agurments like the kube-proxy parameters won't harm other make files.

cmd/main_postsubmit.go
```diff
@@ -0,0 +1,99 @@
...
+		"RELEASE_BRANCH=" + os.Args[2], "RELEASE=" + os.Args[3],
+		"DEVELOPMENT=" + os.Args[4], "AWS_REGION=" + os.Args[5],
+		"AWS_ACCOUNT_ID=" + os.Args[6], "GO_RUNNER_IMAGE=" + os.Args[9],
+		"KUBE_PROXY_BASE_IMAGE=" + os.Args[10], "IMAGE_TAG='$(GIT_TAG)-$(PULL_BASE_SHA)'"}
+	runCommand(gitRoot, cniPluginsChanged, "containernetworking/plugins", os.Args[2], os.Args[3], os.Args[11], buildArg)
```
TerryHowe, line 94: This part becomes something like
```
    cmd := NewCommand()
    cmd.run(cniPluginsChanged, "containernetworking/plugins")
...
```

TerryHowe (changes requested):

cmd/main_postsubmit.go
```diff
@@ -0,0 +1,99 @@
...
+		args[1] = root + "/projects/"
+		if path != "coredns/coredns" {
+			command = exec.Command("bash", root+"/release/lib/create_final_dir.sh", releaseBranch, release, artifactBucket, path)
+			output, err = command.CombinedOutput()
+			if err != nil {
+				log.Fatalf("There was an error running the create_final_dir script: %v. Output:\n%v", err, string(output))
+			}
+			fmt.Printf("Output of the create_final_dir script for %v:\n %v", path, string(output))
+			if path == "kubernetes/kubernetes" {
+				command = exec.Command("/bin/bash", "-c", "mv " + root + "/projects/" + path + "/_output/" + releaseBranch + "/*" + " /logs/artifacts")
```
chanwit, line 31: using Sprintf would make this like more readable.
```
command = exec.Command("/bin/bash", "-c", 
  fmt.Sprintf("mv %s/projects/%s/_output/%s/* /logs/artifacts", root, path, releaseBranch))
```

cmd/main_postsubmit.go
```diff
@@ -0,0 +1,99 @@
...
+			} else {
+				command = exec.Command("/bin/bash", "-c", "mv " + root + "/projects/" + path + "/_output/tar/*" + " /logs/artifacts")
```
chanwit, line 33: Also this one:
```
command = exec.Command("/bin/bash", "-c", 
  fmt.Sprintf("mv %s/projects/%s/_output/tar/* /logs/artifacts", root, path)
```

cmd/main_postsubmit.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		log.Fatalf("There was an error running the git command: %v", err)
+	}
+	gitRoot := strings.Fields(string(gitRootOutput))[0]
+	kubernetesChanged := false
+	coreDnsChanged := false
+	cniPluginsChanged := false
+	iamAuthChanged := false
+	etcdChanged := false
+	gitDiffCommand := []string{"git", "-C", gitRoot, "diff", "--name-only", "HEAD^", "HEAD"}
```
TerryHowe, line 92: Ah, I see, gitRoot was used here. There is probably something clever that could be done about this, but good enough for now.

TerryHowe:
Thanks for getting all those changes in.

TerryHowe (approved):

PR #65, opened 2020-12-07
Docs: Include deployment steps for Libvirt

ccamacho:
Closes #61

*Description of changes:*
This commit adds some initial documentation
for deploying EKS-D on top of a Libvirt host.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #127, opened 2021-01-16
fix: adjust old install steps for new documentation architecture.

ccamacho:
*Description of changes:*

This PR adjust a recent PR merge that includes install steps
for EKS-D on top of a Libvirt host.

Now the installation is aligned with the kops and kubeadm procedures.

### DIFF
--- a/docs/contents/users/install/index.md
+++ b/docs/contents/users/install/index.md
@@ -5,5 +5,6 @@ These instructions are provided by the community and if there are any issues
 with them, please submit a pull request. These installation methods may not
 be tested.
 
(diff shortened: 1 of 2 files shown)


ccamacho:
Hi @TerryHowe thanks for the #65 merge, I just noticed that there are now some more install steps and the Kubeinit's method is not aligned with the current organization you have there, I just moved the .md files and added it to the index. This should be an easy review. Thanks!

TerryHowe (approved):

TerryHowe:
Thanks for the PR! Just what we needed.

Issue #129, opened 2021-01-19
kubelet has broken ECR auth configuration


nairb774:
**What happened**:

Setting `AWS_PROFILE` and `AWS_CONFIG_FILE` in an effort to change the role the kubelet uses to authenticate with AWS services didn't impact how ECR authentication was performed.

**What you expected to happen**:

In an effort to minimize the amount of implicit permissions available to pode on EKS, I moved the kubelet's permissions to a separate role, and instructed the kubelet to assume this new role by updating the relevant systemd unit. Most things seemed to work, but ECR auth didn't seem to be working like everything else. Confused with the odd behavior, as I expected https://github.com/kubernetes/kubernetes/blob/v1.17.12/pkg/credentialprovider/aws/aws_credentials.go#L215-L221 to be the relevant code, I took to decompiling the kubelet available in the provided EKS AMI. To my surprise, there was additional code added which doesn't seem to be documented anywhere: https://github.com/aws/eks-distro/blob/91374b70c163875908715a225fb9bdbb90b10d5b/projects/kubernetes/kubernetes/1-18/patches/0006-EKS-PATCH-using-regional-sts-endpoint-for-assume-ecr.patch#L74-L100 While I haven't yet updated to 1.18, the assembly I found in the Kubelet looks fairly similar to that code. The environment variable `ECR_PULL_ROLE_ARN` seems to be completely unknown to the greater internet (no results come up on Google, Bing or DuckDuckGo), and it would be really appreciated to either have this code properly documented, or ideally, just leave the default credential provider alone so all of the environment variable settings work as one might expect.

**How to reproduce it (as minimally and precisely as possible)**:
Run the following on top of the current managed node group AMI (snippeted from our packer config):

```shell
cat > /etc/systemd/system/kubelet.service.d/aws-profile.conf <<"EOF"
[Service]
# The following config file needs to be created before attempting to start the
# kubelet:
EnvironmentFile=/etc/kubernetes/kubelet/aws.env
EOF
# Small kludge: We need to have the /usr/bin/aws-iam-authenticator (used for
# authenticating the kubelet with the control plane) use the default profile due
# to how the kube-system/aws-auth ConfigMap is managed. This makes a poor
# assumption about the format, but what can you do? We capture the leading
# indentation, and use that for forming the env-var block, hoping that the block
# doesn't exist.
! grep 'env:' /var/lib/kubelet/kubeconfig # Verify assumption
sed -i -e 's/^\( *\)command:.*$/&\
[...]

PR #132, opened 2021-01-20
ecr: Remove regional STS patch preferring updated AWS SDK.

nairb774:
*Issue #, if available:*

https://github.com/aws/eks-distro/issues/129

*Description of changes:*

The patch being removed changed the how the kubelet would locate and
acquire credentials for authenticating against ECR. It would replace the
built in, and default credential location behaviors with a custom path,
and an undocumented set of environment variables. With the patch to
upgrade the kubelet to version 1.34.24 of the AWS SDK, it looks like the
ability to make use of regional STS is possible via the
`sts_regional_endpoints` setting or the `AWS_STS_REGIONAL_ENDPOINTS`
environment variable.

The removal of this patch restores the expected and common credential
location behavior of the AWS SDK and hopefully remove the undocumented
environment variables from being needed.


kschumy:
This is really interesting. I'll look over your issue and this PR as soon as I can

nairb774:
Just want to follow up and see if there is anything else I might need to do.

TerryHowe:
Thanks for this, but it is being implemented in 1.20 currently in the works. https://github.com/aws/eks-distro/pull/227

PR #133, opened 2021-01-21
Parameterized postsubmit arguments

micahhausler:
*Issue #, if available:*
N/A
*Description of changes:*

* Turned postsubmit args into parameterized flags
* Abstracted file changed variables 
* Added `--dry-run` flag

### DIFF
--- a/cmd/main_postsubmit.go
+++ b/cmd/main_postsubmit.go
@@ -1,12 +1,19 @@
 package main
 
 import (
+	"flag"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"os/exec"
-	"strconv"
+	"path/filepath"
 	"strings"
+)
+
+var (
+	outputStream io.Writer = os.Stdout
+	errStream    io.Writer = os.Stderr
 )
 
 type Command struct {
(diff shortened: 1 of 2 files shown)


cmd/main_postsubmit.go
```diff
@@ -1,128 +1,192 @@
...
+	dryRun bool
 }
 
-func (cmd Command) run(runMake bool, uploadArtifacts bool, targetPath string) {
-	if runMake == true {
-		// Add a specific target
-		cmd.args[1] = cmd.args[1] + targetPath
-		command := exec.Command("make", cmd.args...)
-		output, err := command.CombinedOutput()
+func (c *Command) runMake(uploadArtifacts bool, targetPath string) error {
```
bnrjee, line 32: This function tries to make 2 non make exec calls. Probably, we can rename it to only run?

cmd/main_postsubmit.go
```diff
@@ -1,128 +1,192 @@
...
+	}
+	commandArgs = append(commandArgs, c.makeArgs...)
+
+	cmd := exec.Command("make", c.makeArgs...)
+	log.Printf("Executing: %s", strings.Join(cmd.Args, " "))
+	cmd.Stdout = outputStream
+	cmd.Stderr = errStream
+	if !c.dryRun {
+		err := cmd.Run()
 		if err != nil {
```
bnrjee, line 46: Probably we would want to print the output when this call succeeds. Same for the create_final_dir.sh script as well.

cmd/main_postsubmit.go
micahhausler: yea, I can rename it

cmd/main_postsubmit.go
micahhausler: The output of the command is set to forward to the tools stderr/stdout since we don't need to capture/parse it.

bnrjee (approved):

PR #137, opened 2021-01-26
Update the release process

TerryHowe:
Generate development builds that can be tested with our automated conformance tests. This should also make development builds available to the community. The goals here are:

* Make it more gitopsish so that parameters are not buried in Prow
* Making gitopsish will allow any developer to build artifacts locally without special information
* Standardize the Makefiles
* Break down release process so segments can be run as needed

I found it impossible to add a build number to the release, so I had to change RELEASE. Build releases are well out of the way of actual releases.


Makefile
```diff
@@ -1,5 +1,6 @@
 export RELEASE_BRANCH?=1-18
-export RELEASE?=1
+export PULL_NUMBER?=1
```
DanyC97, line 2: @TerryHowe  👋   i would suggest maybe adding few comments and maybe provide a real example so folks who are not AWS employees and may miss internal chats/ context will get it the meaning. 

Otherwise folks will have to go right to the [main script](https://github.com/aws/eks-distro/pull/137/files#diff-cd47c42ec890bee60a3b95dfea693761b7dd991ef47ab10a32ce99aaf32ef08bR18-R20)  .

release/copy_artifacts.sh
```diff
@@ -0,0 +1,35 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euxo pipefail
+
+PROJECT="${1?First required argument is project}"
+RELEASE_BRANCH="${2?Second required argument is release branch for example 1-18}"
+RELEASE="${3?Third required argument is release for example 1}"
```
DanyC97, line 20: as mentioned above i'd presonally provide at the top and in here real example of the release rather than a random number. I hope i'm not seen as being picky, maybe is just me and how i function ;)

Makefile
TerryHowe: Fair point @DanyC97 it makes me think several things
* I hadn't gotten around to even putting a description on this PR yet
*  The `PULL_NUMBER` is really just a build number and probably should not override RELEASE. I think it would be better adding it after the release number like 1.18.1.137
* The pull number is just for build releases. Production releases will go out like 1.18.1
* The pull number environment variable is supplied by Prow and will match the pull request #137 in this case

TerryHowe: The other thing is I think the only internal chats here are in my head :)

DanyC97: thanks for the info @TerryHowe , it does clarify things indeed. 👍

release/copy_artifacts.sh
TerryHowe: I've moved the default releases into gitops style files https://github.com/aws/eks-distro/pull/137/files#diff-0afcd849197deaf1cad81cc32185ee536e2470ccd3d9e9d00513f12aac862e01R1 and the pull number now gets tacked onto the end https://github.com/aws/eks-distro/pull/137/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R7

Makefile
```diff
@@ -1,5 +1,14 @@
+BASE_DIRECTORY=$(shell git rev-parse --show-toplevel)
 export RELEASE_BRANCH?=1-18
-export RELEASE?=1
+DEFAULT_RELEASE=$(shell cat $(BASE_DIRECTORY)/release/$(RELEASE_BRANCH)/RELEASE)
+ifneq ("$(PULL_NUMBER)","")
+	export RELEASE?=$(shell expr $(DEFAULT_RELEASE) \* 10000 + $(PULL_NUMBER))
```
vivek-koppuru, line 5: What are we trying to do here?

Makefile
```diff
@@ -1,5 +1,14 @@
...
+else
+	export RELEASE?=$(DEFAULT_RELEASE)
+endif
+ARTIFACT_BUCKET?=$(shell cat $(BASE_DIRECTORY)/release/ARTIFACT_BUCKET)
```
vivek-koppuru, line 9: Instead of having individual text files for stuff like this, does it make sense to store and read from a single file? I think we definitely want to try to be conservative with the amount of files we create and limit it potentially to the GIT_TAG for now.

projects/coredns/coredns/Makefile
```diff
@@ -1,21 +1,17 @@
...
+	export RELEASE?=$(shell expr $(DEFAULT_RELEASE) \* 10000 + $(PULL_NUMBER))
+else
+	export RELEASE?=$(DEFAULT_RELEASE)
 endif
+ARTIFACT_BUCKET?=$(shell cat $(BASE_DIRECTORY)/release/ARTIFACT_BUCKET)
+GIT_TAG?=$(shell cat GIT_TAG)
 
-RELEASE_BRANCH?="1-18"
-RELEASE?="1"
+REPO=coredns
```
vivek-koppuru, line 12: We are removing the validations and making them hard values now? Is it because you don't see us having to change those values?

release/ARTIFACT_BUCKET
```diff
@@ -0,0 +1 @@
+prowpresubmitsdataclusterstack-prowbucket7c73355c-vfwwxd2eb4gp
```
vivek-koppuru, line 1: Yea I feel that since this is not something we would actively change like RELEASE or GIT_TAG, wondering if we can store in a common constants file or something. Also would rather do what is considered "best practice" with regards to individual files vs a constants file.

release/s3_sync.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+# Copyright 2021 Amazon.com Inc. or its affiliates. All Rights Reserved.
```
vivek-koppuru, line 2: I think we decided that we don't actually need the year, so we should just start removing the year from our files so that it won't change on a yearly basis.

Makefile
TerryHowe: This is generating a release number for a development build. In this case it would be 10137 since this is PR #137

Makefile
TerryHowe: It is probably a waste of time to get the artifact bucket out of the repo, but the release numbers are a good thing. I'm thinking we'll trigger new releases just by changing those release files.

projects/coredns/coredns/Makefile
TerryHowe: I think making everything configurable is unwise. This will always just build coredns, nothing else

release/s3_sync.sh
TerryHowe: Fair enough, we should have one PR to remove them. Please open a ticket if this is a thing.

Makefile
TerryHowe: I removed that attifact backet file. That is better handled in prow.

TerryHowe:
Waiting on https://github.com/aws/eks-distro-prow-jobs/pull/87

Makefile
vivek-koppuru: Why not make this the default for development? Is there a reason this can't just be configurable instead of forcing the user to use this logic?

Makefile
```diff
@@ -50,9 +57,13 @@ postsubmit-conformance:
 		--dry-run=false
 	bash development/kops/prow.sh
 
+.PHONY: upload
+upload:
+	release/s3_sync.sh $(RELEASE_BRANCH) $(RELEASE) $(ARTIFACT_BUCKET)
```
vivek-koppuru, line 62: If we are passing the ARTIFACT_BUCKET here anyways, let's not export it above then.

projects/containernetworking/plugins/Makefile
```diff
@@ -18,6 +27,8 @@ build: tarballs
 
 .PHONY: release
 release: tarballs
+	$(BASE_DIRECTORY)/release/copy_artifacts.sh $(COMPONENT) $(RELEASE_BRANCH) $(RELEASE)
```
vivek-koppuru, line 30: Same here, anything we pass in to the scripts, we should not export above. Exporting means that our child bash scripts would have access to the environment variable too I believe.

projects/coredns/coredns/Makefile
vivek-koppuru: Makes sense

release/s3_sync.sh
vivek-koppuru: Yea at least for this PR and going forward, let's remove the dates. We can open a ticket for the ones that already exist in the future, though it doesn't matter too much for ones we committed already.

Makefile
TerryHowe: This isn't forcing the user to do anything, this is for Prow. PULL_NUMBER is a prow environment variable, although we are going to have to set it manually for postsubmits.

projects/containernetworking/plugins/Makefile
TerryHowe: The export of RELEASE_BRANCH was from a previous PR and I agree it isn't too cool. My survey of our shell files seems to indicate it will work without exporting that value. It is passed in all the scripts I could find that mentioned the variable.

Makefile
TerryHowe: I wrote a little test Makefile setup with a subdirectory and a shell script and exporting add passing these arguments seems to be unnecessary

Makefile
```diff
@@ -1,16 +1,23 @@
...
-export RELEASE?=1
-export DEVELOPMENT?=false
-export AWS_ACCOUNT_ID?=$(shell aws sts get-caller-identity --query Account --output text)
-export AWS_REGION?=us-west-2
-export IMAGE_REPO?=$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com
+BASE_DIRECTORY=$(shell git rev-parse --show-toplevel)
+RELEASE_BRANCH?=1-18
+DEFAULT_RELEASE=$(shell cat $(BASE_DIRECTORY)/release/$(RELEASE_BRANCH)/RELEASE)
+ifneq ("$(PULL_NUMBER)","")
+	RELEASE?=$(shell expr $(DEFAULT_RELEASE) \* 10000 + $(PULL_NUMBER))
```
vivek-koppuru, line 5: As discussed offline, let's use commit hash instead of this being dependent on the PR number so that it is cleaner and more logical.

vivek-koppuru (approved):
/lgtm

PR #139, opened 2021-01-29
Run kops and copy junit file

bnrjee:
*Issue #, if available:*

*Description of changes:*


Makefile
```diff
@@ -4,12 +4,14 @@ export DEVELOPMENT?=false
...
 export IMAGE_REPO?=$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com
+export KOPS_CLUSTER_NAME=$(shell cat /dev/urandom | LC_ALL=C tr -dc '[:alpha:]' | tr '[:upper:]' '[:lower:]' | head -c 32).prod.test.pdx.kops-ci.model-rocket.aws.dev
+export TEST_ROLE_ARN=arn:aws:iam::$(AWS_ACCOUNT_ID):role/$(TEST_ROLE_NAME)
 BASE_IMAGE_TAG?=$(shell cat EKS_DISTRO_BASE_TAG_FILE)
 export BASE_IMAGE?=$(IMAGE_REPO)/eks-distro/base:$(BASE_IMAGE_TAG)
 KUBE_BASE_TAG?=v0.4.2-ea45689a0da457711b15fa1245338cd0b636ad4b
 export KUBE_PROXY_BASE_IMAGE?=$(IMAGE_REPO)/kubernetes/kube-proxy-base:$(KUBE_BASE_TAG)
 export GO_RUNNER_IMAGE?=$(IMAGE_REPO)/kubernetes/go-runner:$(KUBE_BASE_TAG)
-ARTIFACT_BUCKET?=my-s3-bucket
+ARTIFACT_BUCKET?=artifactsstack-3794122512-artifactsbucket2aac5544-1f3dgu9wrpiz2
```
vivek-koppuru, line 14: I would rather this value be in the prowjob and not on the Makefile that the customer runs as well.

Makefile
```diff
@@ -4,12 +4,14 @@ export DEVELOPMENT?=false
 export AWS_ACCOUNT_ID?=$(shell aws sts get-caller-identity --query Account --output text)
 export AWS_REGION?=us-west-2
 export IMAGE_REPO?=$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com
+export KOPS_CLUSTER_NAME=$(shell cat /dev/urandom | LC_ALL=C tr -dc '[:alpha:]' | tr '[:upper:]' '[:lower:]' | head -c 32).prod.test.pdx.kops-ci.model-rocket.aws.dev
+export TEST_ROLE_ARN=arn:aws:iam::$(AWS_ACCOUNT_ID):role/$(TEST_ROLE_NAME)
```
vivek-koppuru, line 8: This account id will not be correct then right? It would use build account rather than test-build account

development/kops/auth.sh
```diff
@@ -0,0 +1,12 @@
+unset AWS_ACCESS_KEY_ID
+unset AWS_SECRET_ACCESS_KEY
+unset AWS_SESSION_TOKEN
+export AWS_DEFAULT_PROFILE=default
+export AWS_PROFILE=default
+assume_test_role_output=`aws sts assume-role --role-arn $TEST_ROLE_ARN --role-session-name test-role-session`
+export AWS_ACCESS_KEY_ID=`echo $assume_test_role_output|jq -r .Credentials.AccessKeyId`
```
vivek-koppuru, line 7: I thought we were trying to avoid doing this and instead use config file no? Based on what Micah showed us about how it might not be an SDK issue, unless we concluded that it's still a Kops issue.

vivek-koppuru (changes requested):

development/kops/auth.sh
bnrjee: So far i could not get it to work with prow.

Makefile
```diff
@@ -4,12 +4,13 @@ export DEVELOPMENT?=false
 export AWS_ACCOUNT_ID?=$(shell aws sts get-caller-identity --query Account --output text)
 export AWS_REGION?=us-west-2
 export IMAGE_REPO?=$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com
+export KOPS_CLUSTER_NAME=$(shell cat /dev/urandom | LC_ALL=C tr -dc '[:alpha:]' | tr '[:upper:]' '[:lower:]' | head -c 32).prod.test.pdx.kops-ci.model-rocket.aws.dev
```
vivek-koppuru, line 7: Can we use postsubmit job name to define our cluster name?

Makefile
```diff
@@ -4,12 +4,13 @@ export DEVELOPMENT?=false
...
 BASE_IMAGE_TAG?=$(shell cat EKS_DISTRO_BASE_TAG_FILE)
 export BASE_IMAGE?=$(IMAGE_REPO)/eks-distro/base:$(BASE_IMAGE_TAG)
 KUBE_BASE_TAG?=v0.4.2-ea45689a0da457711b15fa1245338cd0b636ad4b
 export KUBE_PROXY_BASE_IMAGE?=$(IMAGE_REPO)/kubernetes/kube-proxy-base:$(KUBE_BASE_TAG)
 export GO_RUNNER_IMAGE?=$(IMAGE_REPO)/kubernetes/go-runner:$(KUBE_BASE_TAG)
-ARTIFACT_BUCKET?=my-s3-bucket
+export ARTIFACT_BUCKET?=my-s3-bucket
```
vivek-koppuru, line 13: Will the export here actually change anything? Or it would actually persist in the script?

development/kops/prow.sh
```diff
@@ -13,33 +13,32 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-if [ "$PROW_ROLE_ARN" == "" ]; then
-    echo "Empty PROW_ROLE_ARN, this script must be run in a postsubmit pod with IAM Roles for Service Accounts"
+if [ "$AWS_ROLE_ARN" == "" ]; then
```
vivek-koppuru, line 16: Maybe call this SOURCE_ROLE_ARN?

development/kops/run_all.sh
```diff
@@ -13,18 +13,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -exo pipefail
```
vivek-koppuru, line 16: Ideally we wouldn't have to remove `x`

development/kops/run_sonobuoy.sh
```diff
@@ -58,6 +58,6 @@ mv $results "./${KOPS_CLUSTER_NAME}/$results"
 results="./${KOPS_CLUSTER_NAME}/$results"
 mkdir ./${KOPS_CLUSTER_NAME}/results
 tar xzf $results -C ./${KOPS_CLUSTER_NAME}/results
-cp ./${KOPS_CLUSTER_NAME}/results/plugins/e2e/results/global/junit_01.xml .
+cp ./${KOPS_CLUSTER_NAME}/results/plugins/e2e/results/global/junit_01.xml /logs/artifacts
```
vivek-koppuru, line 61: We would want to figure out adding a Spyglass lens here to display the junit output

development/kops/run_all.sh
TerryHowe: The -x option risks deplaying credentials in a public log. We might want to leave the rest though.

Makefile
bnrjee: We are generating the s3 bucket name. Wanted to follow the same logic here.

development/kops/prow.sh
bnrjee: I just used the same name we are using for docs.

Makefile
bnrjee: It does get propagated to the down level scripts, if that is what you are asking.

development/kops/run_sonobuoy.sh
bnrjee: I think, we already have this,
https://github.com/aws/eks-distro-build-tooling/blob/4ad4032bd9fd457853a1d8af7788a[…]harts/stable/prow-control-plane/templates/config-ConfigMap.yaml

bnrjee: https://github.com/aws/eks-distro-build-tooling/blob/4ad4032bd9fd457853a1d8af7788a1d36de9dc3c/helm-charts/stable/prow-control-plane/templates/config-ConfigMap.yaml#L59

Makefile
TerryHowe: maybe we could move this into create_store.sh and use the same number for bucket and domain name

TerryHowe: I don't think that is important right now though.

development/kops/run_all.sh
vivek-koppuru: But with the latest revision we don't have that risk right?

Makefile
vivek-koppuru: I think it would be more beneficial for debugging purposes, and even for the s3 bucket name if we are generating it on every run and deleting the bucket, then we should just keep the job id in there.

micahhausler: Can we just use the same S3 bucket for all clusters? We can only crate 100 per account and I'd hate to have bucket leakage prevent new clusters from getting created

micahhausler (changes requested):

development/kops/install.sh
```diff
@@ -0,0 +1,6 @@
+set -eo pipefail
+curl -LO https://github.com/kubernetes/kops/releases/download/v1.18.2/kops-linux-amd64
+chmod +x kops-linux-amd64
+mv kops-linux-amd64 /usr/local/bin/kops
+curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
```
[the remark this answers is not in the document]
TerryHowe: https://distro.eks.amazonaws.com/kubernetes-1-18/releases/1/artifacts/kubernetes/v1.18.9/bin/linux/arm64/kubectl

Makefile
jaxesn: Random thought, should we just use a local path for the state store since these clusters are shortlived?  One downside of local storage is if things fail to clean up, we could get the state from s3 to do a manual clean up.  That said, we may want to add a failsafe process for clean up to avoid having to do anything manually.  Maybe like a nightly cron job to "reset" the aws account.

TerryHowe: kops is all set up to use s3.

vivek-koppuru:
Closing this PR in favor of https://github.com/aws/eks-distro/pull/146

PR #141, opened 2021-01-29
Better BuildKit practices

TerryHowe:
Couple improvements to BuildKit development usage:

* Stop using the development build
* Add README instructions so I can find them easier
* Rename the file to something less generic

### DIFF
--- a/development/buildkit-compose.yaml
+++ b/development/buildkit-compose.yaml
@@ -15,7 +15,7 @@
 version: "3.8"
 services:
   buildkitd:
-    image: moby/buildkit:master-rootless
+    image: moby/buildkit:v0.8.1-rootless
     command:
     - --addr
     - tcp://0.0.0.0:1234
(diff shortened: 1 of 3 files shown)


development/buildkit-compose.yaml
```diff
@@ -15,7 +15,7 @@
 version: "3.8"
 services:
   buildkitd:
-    image: moby/buildkit:master-rootless
+    image: moby/buildkit:v0.8.1-rootless
```
abhay-krishna, line 18: What was the reason for moving from `master` tag to `v0.8.1`? Is it because the `master` image is updated often and might include breaking changes? Is this something that you recommend doing on our Prowjobs as well?

TerryHowe: master is the dev build off the head of the master branch. Yeh, we should probably do that for Prow.

abhay-krishna: Gotcha, so we would want to use the latest stable version from DockerHub instead of dev builds.

abhay-krishna (approved):
/lgtm
/approve

PR #143, opened 2021-02-02
Add build target to CNI make file

TerryHowe:
Add build target to CNI make file, so we can change the target for Prow presubmit to what it should use.

### DIFF
--- a/projects/containernetworking/plugins/Makefile
+++ b/projects/containernetworking/plugins/Makefile
@@ -1,18 +1,8 @@
-REPO?=plugins
-CLONE_URL?=https://github.com/containernetworking/$(REPO).git
 GIT_TAG?=$(shell cat GIT_TAG)
 
-ifeq ("$(REPO)","")
-	$(error No repository name was provided.)
-endif
-
-ifeq ("$(CLONE_URL)","")
-	$(error No clone url was provided.)
-endif
-
-ifeq ("$(GIT_TAG)","")
-	$(error No git tag was provided.)
-endif
+REPO=plugins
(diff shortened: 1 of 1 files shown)


projects/containernetworking/plugins/Makefile
```diff
@@ -22,8 +12,13 @@ binaries:
 tarballs:  binaries
 	build/create_tarballs.sh $(REPO) $(GIT_TAG)
 
+.PHONY: build
+build: tarballs
+	echo "Done $(COMPONENT)"
+
 .PHONY: release
 release: tarballs
+	echo "Done $(COMPONENT)"
```
vivek-koppuru, line 21: So the release target is not uploading to S3 vs build? What is the situation there?

TerryHowe: That is the change I put in on #137 which hasn't merged yet

jaxesn: to confirm, you are saying once 137 is merged, the release target will upload after build?  Thats how it looks [here](https://github.com/aws/eks-distro/pull/137/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R66)

vivek-koppuru (approved):
/lgtm

projects/containernetworking/plugins/Makefile
TerryHowe: This change is just adding a build target to the Make file so the presubmit tests will pass for #137 The #137 PR adds the upload to s3 like you pointed out in your link

PR #145, opened 2021-02-03
Add kube-proxy patch to allow change of base image

TerryHowe:
Add kube-proxy patch to 1-19


projects/kubernetes/kubernetes/1-19/patches/0012-EKS-PATCH-allow-override-of-kube-proxy-base-image
```diff
@@ -0,0 +1,30 @@
+diff --git a/build/common.sh b/build/common.sh
```
vivek-koppuru, line 1: Are there any sign off notes we want to include here, or are we just not including them anymore?

TerryHowe: Notes added.

TerryHowe:
I still need to investigate if our kube-proxy base image meets all the requirements expected, but this is good to go either way. If we need a change with the kube-proxy base, I'll make that in another PR.

TerryHowe:
We already have logic to use AL2 for kube-proxy-base and we run `yum update` on it.

abhay-krishna (approved):
/lgtm
/approve

PR #146, opened 2021-02-03
Add build logic and initial kops files

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

Continuation of Abhisek's PR here: https://github.com/aws/eks-distro/pull/139

We will be adding all the logic to build and upload the artifacts with our current logic for the scope of this PR. After this, we need to create the necessary IAM roles and S3 buckets for kops to consume in those scripts. Then we will uncomment the kops script and test that logic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ BASE_IMAGE?=$(IMAGE_REPO)/eks-distro/base:$(BASE_IMAGE_TAG)
 KUBE_BASE_TAG?=v0.4.2-ea45689a0da457711b15fa1245338cd0b636ad4b
 KUBE_PROXY_BASE_IMAGE?=$(IMAGE_REPO)/kubernetes/kube-proxy-base:$(KUBE_BASE_TAG)
 GO_RUNNER_IMAGE?=$(IMAGE_REPO)/kubernetes/go-runner:$(KUBE_BASE_TAG)
+RELEASE_AWS_PROFILE?=default
 
 ifdef MAKECMDGOALS
 TARGET=$(MAKECMDGOALS)
@@ -35,7 +36,20 @@ setup:
 	AWS_DEFAULT_PROFILE=$(RELEASE_AWS_PROFILE) bash ./ecr-public/get-credentials.sh
 
 .PHONY: build
-build: makes
+build:
+	go vet cmd/main_postsubmit.go
+	go run cmd/main_postsubmit.go \
+		--target=build \
+		--release-branch=${RELEASE_BRANCH} \
+		--release=${RELEASE} \
+		--development=${DEVELOPMENT} \
+		--region=${AWS_REGION} \
+		--account-id=${AWS_ACCOUNT_ID} \
+		--base-image=${BASE_IMAGE} \
+		--image-repo=${IMAGE_REPO} \
+		--go-runner-image=${GO_RUNNER_IMAGE} \
+		--kube-proxy-base=${KUBE_PROXY_BASE_IMAGE} \
+		--dry-run=true
 	@echo 'Done' $(TARGET)
 
 .PHONY: postsubmit-conformance
@@ -53,9 +67,8 @@ postsubmit-conformance:
 		--go-runner-image=${GO_RUNNER_IMAGE} \
 		--kube-proxy-base=${KUBE_PROXY_BASE_IMAGE} \
 		--artifact-bucket=$(ARTIFACT_BUCKET) \
-		--upload-to-s3=false \
 		--dry-run=false
-	bash development/kops/prow.sh
+#	bash development/kops/prow.sh
 
 .PHONY: upload
 upload:
--- a/cmd/main_postsubmit.go
+++ b/cmd/main_postsubmit.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 )
 
@@ -21,7 +22,6 @@ type Command struct {
 	gitRoot              string
 	release              string
 	artifactBucket       string
-	uploadToPublicBucket bool
 
 	makeTarget string
 	makeArgs   []string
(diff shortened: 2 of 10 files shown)


Makefile
```diff
@@ -1,5 +1,6 @@
 BASE_DIRECTORY=$(shell git rev-parse --show-toplevel)
 RELEASE_BRANCH?=1-18
+RELEASE?=1
```
TerryHowe, line 3: not terribly important, but that ^ should not be there

Makefile
```diff
@@ -13,11 +14,13 @@ DEVELOPMENT?=false
 AWS_ACCOUNT_ID?=$(shell aws sts get-caller-identity --query Account --output text)
 AWS_REGION?=us-west-2
 IMAGE_REPO?=$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com
+export KOPS_CLUSTER_NAME=$(shell cat /dev/urandom | LC_ALL=C tr -dc '[:alpha:]' | tr '[:upper:]' '[:lower:]' | head -c 32).prod.test.pdx.kops-ci.model-rocket.aws.dev
```
TerryHowe, line 17: Delete this  and put it in the create_store_name script most likely

development/kops/run_all.sh
```diff
@@ -22,6 +22,7 @@ source ./set_environment.sh
 $PREFLIGHT_CHECK_PASSED || exit 1
 ./download_kops.sh
 ./create_values_yaml.sh
+./install.sh
```
TerryHowe, line 25: Delete this ^

development/kops/install.sh
```diff
@@ -0,0 +1,6 @@
+set -eo pipefail
+curl -LO https://github.com/kubernetes/kops/releases/download/v1.18.2/kops-linux-amd64
+chmod +x kops-linux-amd64
+mv kops-linux-amd64 /usr/local/bin/kops
+curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
```
TerryHowe, line 6: Remove this file this is covered by download_kops.sh

development/kops/create_configuration.sh
```diff
@@ -53,9 +53,13 @@ echo "Creating cluster configuration"
 ${KOPS} create -f "./${KOPS_CLUSTER_NAME}/${KOPS_CLUSTER_NAME}.yaml"
 
 echo "Creating cluster ssh key"
-export SSH_KEY_PATH=${SSH_KEY_PATH:-$HOME/.ssh/id_rsa.pub}
-${KOPS} create secret --name $KOPS_CLUSTER_NAME sshpublickey admin -i ${SSH_KEY_PATH}
```
TerryHowe, line 57: This should not have been deleted The command is ${KOPS}

Makefile
TerryHowe: If we put it there, we can generate a store name and cluster name the same way using the pull sha1

development/kops/install.sh
abhay-krishna: We still need `kubectl` right? Can we just install it in-situ in the script where we use it? Or include in the documentation that users must have kubectl installed before running the kops scripts?

Also I think we should use the build of kubectl that EKS-D vends - `https://distro.eks.amazonaws.com/kubernetes-1-18/releases/1/artifacts/kubernetes/v1.18.9/bin/linux/amd64/kubectl`. Templatizing the kubectl version in the URL is a good idea too.

TerryHowe: Yeh, good call on kubectl

Makefile
vivek-koppuru: Yup that's what I want to do, just doing some testing before I address that

development/kops/create_configuration.sh
vivek-koppuru: Good catch, I still think we wanted to generate a random ssh key but I can make that change below

Makefile
```diff
@@ -36,6 +38,21 @@ setup:
...
+		--release=${RELEASE} \
+		--development=${DEVELOPMENT} \
+		--region=${AWS_REGION} \
+		--account-id=${AWS_ACCOUNT_ID} \
+		--base-image=${BASE_IMAGE} \
+		--image-repo=${IMAGE_REPO} \
+		--go-runner-image=${GO_RUNNER_IMAGE} \
+		--kube-proxy-base=${KUBE_PROXY_BASE_IMAGE} \
+		--artifact-bucket=$(ARTIFACT_BUCKET) \
+		--upload-to-s3=true \
```
abhay-krishna, line 54: Will this upload step not fail in the `s3_sync` script, since we do not have a bucket of that name in the account?

cmd/main_postsubmit.go
```diff
@@ -61,32 +59,28 @@ func (c *Command) buildProject(projectPath string, uploadArtifacts bool) error {
...
 			cmd.Stderr = errStream
 			log.Printf("Executing: %s", strings.Join(cmd.Args, " "))
-			if !c.dryRun {
-				err := cmd.Run()
-				if err != nil {
-					return fmt.Errorf("Error running create_final_dir.sh: %v", err)
-				}
+			err := cmd.Run()
+			if err != nil {
+				return fmt.Errorf("Error running create_final_dir.sh: %v", err)
```
abhay-krishna, line 64: Should we be calling `create_s3_bucket` and `s3_sync` scripts here instead?

Makefile
TerryHowe: the prow job should be passing in the postsubmit bucket name. I don't recall what bucket that is.

development/kops/install.sh
```diff
@@ -0,0 +1,8 @@
+set -eo pipefail
+curl -LO https://github.com/kubernetes/kops/releases/download/v1.18.2/kops-linux-amd64
```
TerryHowe, line 2: I *think* we don't need this curl and this file has no copyright. I think you meant to delete this file.

vivek-koppuru: Oh yes I wanted to delete this file, I'll do that in my next revision

TerryHowe (approved):

PR #148, opened 2021-02-04
Copy CRDs one at a time

TerryHowe:
Copy CRD into bucket

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ postsubmit-conformance:
 .PHONY: upload
 upload:
 	release/generate_crd.sh $(RELEASE_BRANCH) $(RELEASE)
-	release/s3_sync.sh $(RELEASE_BRANCH) $(RELEASE) $(ARTIFACT_BUCKET) ""
+	release/s3_sync.sh $(RELEASE_BRANCH) $(RELEASE) $(ARTIFACT_BUCKET)
 	@echo 'Done' $(TARGET)
 
 .PHONY: release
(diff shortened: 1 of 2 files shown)


release/s3_sync.sh
```diff
@@ -28,9 +28,19 @@ else
   PUBLIC_READ=''
 fi
 cd ${BASE_DIRECTORY}
-DEST_DIR=kubernetes-${RELEASE_BRANCH}/releases/${RELEASE}/artifacts
-if [ ! -z "${REPO}" ]
+BASE_DIR=kubernetes-${RELEASE_BRANCH}
```
abhay-krishna, line 31: Can we name this something along the lines of PREFIX_DIR or RELEASE_BASE_DIR?

Makefile
```diff
@@ -73,7 +73,7 @@ postsubmit-conformance:
 .PHONY: upload
 upload:
 	release/generate_crd.sh $(RELEASE_BRANCH) $(RELEASE)
-	release/s3_sync.sh $(RELEASE_BRANCH) $(RELEASE) $(ARTIFACT_BUCKET)
+	release/s3_sync.sh $(RELEASE_BRANCH) $(RELEASE) $(ARTIFACT_BUCKET) ""
```
abhay-krishna, line 76: I think it works even if you don't pass in the quotes here. As far as I have seen, Bash automatically takes missing arguments to be empty strings.

release/s3_sync.sh
TerryHowe: Done

abhay-krishna (approved):
/lgtm
/approve

PR #151, opened 2021-02-04
Fix make file for postsubmit

TerryHowe:
Make file broken for postsubmit. I'll try something else to fix it for normal devs.


jaxesn (approved):
/lgtm

PR #152, opened 2021-02-04
Set objects to public if not development, fix image tags for builds, …

vivek-koppuru:
…and order our builds

*Issue #, if available:*

*Description of changes:*

1. Setting the objects to public read based on the DEVELOPMENT env var
2. Fixing the image tags for builds to make it more consistent. This way, we don't need to pass in the name of the images from the main makefile, and rather make use of the defaults which should be logical for our builds.
3. Order the builds so that the kubernetes/release happens first so that the kubernetes/kubernetes build can use the images built by kubernetes/release.


Makefile
```diff
@@ -7,17 +7,11 @@ else
...
 
 DEVELOPMENT?=false
 AWS_ACCOUNT_ID?=$(shell aws sts get-caller-identity --query Account --output text)
 AWS_REGION?=us-west-2
 IMAGE_REPO?=$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com
-BASE_IMAGE_TAG?=$(shell cat EKS_DISTRO_BASE_TAG_FILE)
-BASE_IMAGE?=$(IMAGE_REPO)/eks-distro/base:$(BASE_IMAGE_TAG)
-KUBE_BASE_TAG?=v0.4.2-ea45689a0da457711b15fa1245338cd0b636ad4b
-KUBE_PROXY_BASE_IMAGE?=$(IMAGE_REPO)/kubernetes/kube-proxy-base:$(KUBE_BASE_TAG)
-GO_RUNNER_IMAGE?=$(IMAGE_REPO)/kubernetes/go-runner:$(KUBE_BASE_TAG)
```
TerryHowe, line 20: Great, I wanted to get rid of that stuff!

TerryHowe (approved):
/lgtm

TerryHowe (approved):
/lgtm

PR #153, opened 2021-02-05
Change cleanup of kops

TerryHowe:
Change the kops cleanup script so it deletes things (except the bucket) on interrupt or exit.


vivek-koppuru (approved):
/lgtm

PR #154, opened 2021-02-05
Reorganize make target dependencies in Kubernetes due to failed jobs

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

Builds in prow are failing due to being unable to find the kubernetes/kubernetes/kubernetes folder. Reorganizing the target dependencies to ensure that we have a streamlined structure for both our presubmits and postsubmits.

Removing other unnecessary environment variables as our Makefile will construct those values based on the RELEASE. My assumption before was that if they were empty anyways, they would be set in the Makefile but that wasn't the case.


TerryHowe (approved):

jaxesn (approved):
/lgtm

TerryHowe:
We cool?

jaxesn (approved):
/lgtm

PR #155, opened 2021-02-05
Use BUILD_ID instead of sha1 so CRD can be generated

TerryHowe:
The CRD generation insists on a number rather than a string for release, so use BUILD_ID instead of SHA1 for development builds.


jaxesn:
What are the requirements on the CRD side?  Just that it is an integer?  I think these build numbers get pretty big so just making sure there isnt some kind of limit.

jaxesn (approved):
/lgtm

PR #156, opened 2021-02-05
Add local image target to k8s make file

TerryHowe:
We need the tar balls and other associated files for release, so call the local images target for release.


jaxesn (approved):
/lgtm

PR #157, opened 2021-02-05
Remove references to PULL_NUMBER

TerryHowe:
We aren't going to use PULL_NUMBER for release. It will either be sha1 or build number. Either way, it will be passed in as RELEASE


jaxesn (approved):
/lgtm

PR #158, opened 2021-02-05
Kops override artifact repos and remove PULL_NUMBER as release

TerryHowe:
Two things here:

1) Get rid of PULL_NUMBER the Prow environment variable from determining release. The calling script should set and export RELEASE
2) Allow overridiing the artifact web server and ECR repo.


development/kops/set_environment.sh
```diff
@@ -70,6 +64,7 @@ export PREFLIGHT_CHECK_PASSED
 VERSION=$(cat ${BASEDIR}/../../projects/kubernetes/kubernetes/${RELEASE_BRANCH}/GIT_TAG)
```
abhay-krishna, line 64: Is this value being used in any script? Should it be exported in that case?

TerryHowe: Done. It was working because VERSION was still in scope, but better to export for consistency

jaxesn (approved):
/lgtm

jaxesn (approved):
/lgtm

PR #164, opened 2021-02-09
Fix 1-18 Kops Tests

TerryHowe:
Fix 1-18 kOps tests

### DIFF
--- a/development/kops/create_values_yaml.sh
+++ b/development/kops/create_values_yaml.sh
@@ -85,8 +85,8 @@ kubernetesVersion: ${ARTIFACT_URL}/kubernetes-${RELEASE_BRANCH}/releases/${RELEA
 clusterName: $KOPS_CLUSTER_NAME
 configBase: $KOPS_STATE_STORE/$KOPS_CLUSTER_NAME
 awsRegion: $AWS_DEFAULT_REGION
-controlPlaneInstanceProfileArn: ''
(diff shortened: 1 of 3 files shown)


development/kops/create_values_yaml.sh
```diff
@@ -85,6 +85,8 @@ kubernetesVersion: ${ARTIFACT_URL}/kubernetes-${RELEASE_BRANCH}/releases/${RELEA
 clusterName: $KOPS_CLUSTER_NAME
 configBase: $KOPS_STATE_STORE/$KOPS_CLUSTER_NAME
 awsRegion: $AWS_DEFAULT_REGION
+controlPlaneInstanceProfileArn: ''
```
abhay-krishna, line 88: Wondering if this will lead to duplicate fields in values.yaml if the profile ARN's are set as environment variables 🤔

TerryHowe: Good point, I forgot that stuff was down there

TerryHowe: Fixed

jaxesn:
Why do we want to do this?  Shouldnt we only set them if they are supplied?

TerryHowe:
The template did not have these values and so create configuration failed if they were not set.

TerryHowe:
1-18 tests are passing again after this.

jaxesn (approved):
/lgtm

PR #168, opened 2021-02-09
Release prow script

TerryHowe:

### DIFF
--- a/release/prow.sh
+++ b/release/prow.sh
@@ -35,7 +35,7 @@ web_identity_token_file=/var/run/secrets/eks.amazonaws.com/serviceaccount/token
 
 [profile release-account]
 role_arn=$ARTIFACT_DEPLOYMENT_ROLE_ARN
-region=${AWS_REGION:-${AWS_DEFAULT_REGION:-us-west-2}}
+region=${AWS_REGION:-${AWS_DEFAULT_REGION:-us-east-1}}
 source_profile=default
 EOF
 export AWS_CONFIG_FILE=$(pwd)/awscliconfig
--- a/release/prow.sh
+++ b/release/prow.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+# Copyright 2020 Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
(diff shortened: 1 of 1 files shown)


release/prow.sh
```diff
@@ -0,0 +1,43 @@
...
+cat << EOF > awscliconfig
+[default]
+output=json
+region=${AWS_REGION:-${AWS_DEFAULT_REGION:-us-west-2}}
+role_arn=$AWS_ROLE_ARN
+web_identity_token_file=/var/run/secrets/eks.amazonaws.com/serviceaccount/token
+
+[profile release-account]
+role_arn=$ARTIFACT_DEPLOYMENT_ROLE_ARN
+region=${AWS_REGION:-${AWS_DEFAULT_REGION:-us-west-2}}
```
abhay-krishna, line 38: I'm thinking we should hardcode region to `us-east-1` here, since that is where the artifacts account is located.

TerryHowe: I'd be reluctant to change anything at this point. We normally run in us-west-2 and ECR public commands specify east if they need to already. S3 is region free.

TerryHowe: I made that us-east-1

abhay-krishna (approved):
/lgtm
/approve

release/prow.sh
```diff
@@ -0,0 +1,43 @@
...
+cat << EOF > awscliconfig
+[default]
+output=json
+region=${AWS_REGION:-${AWS_DEFAULT_REGION:-us-west-2}}
+role_arn=$AWS_ROLE_ARN
+web_identity_token_file=/var/run/secrets/eks.amazonaws.com/serviceaccount/token
+
+[profile release-account]
+role_arn=$ARTIFACT_DEPLOYMENT_ROLE_ARN
+region=${AWS_REGION:-${AWS_DEFAULT_REGION:-us-east-1}}
```
vivek-koppuru, line 38: Using AWS_REGION will be a problem here as the Pod already injects the region. Maybe we can add an `ARTIFACTS_AWS_REGION` env var to our prowjob and use that here.

vivek-koppuru (changes requested):

PR #171, opened 2021-02-10
ECR repository commands

TerryHowe:
Standardize the ECR repository commands for deployment


vivek-koppuru (commented):
Does it make sense now to not overuse the `development` directory for all this? Wondering if this a place for `build` instead. Also wondering if it makes sense to have a separate target for creating ECR public repos, and then if customers want to replicate, that would be a prereq for them. And then when public ecr endpoint is passed in for the `make release`, then we should be able to figure out whether we set up config regularly or for ecr-public

micahhausler (approved):
/lgtm

micahhausler:
/hold 
unhold if you wanted to make any more changes

TerryHowe:
> Does it make sense now to not overuse the `development` directory for all this? Wondering if this a place for `build` instead. Also wondering if it makes sense to have a separate target for creating ECR public repos, and then if customers want to replicate, that would be a prereq for them. And then when public ecr endpoint is passed in for the `make release`, then we should be able to figure out whether we set up config regularly or for ecr-public

The build directory seems pretty focused on building. I originally had these in the release directory, but I think they are needed by regular developers to make repos they can push to, so I moved them here. At least they are in a subdirectory. If anything, maybe things like kops should be in another directory.

PR #175, opened 2021-02-11
Remove create all ECR repos from make

TerryHowe:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ presubmit-cleanup = \
 
 .PHONY: setup
 setup:
-	bash ./development/ecr/ecr-command.sh install-ecr-public
+	development/ecr/ecr-command.sh install-ecr-public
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -25,7 +25,6 @@ presubmit-cleanup = \
 setup:
 	bash ./development/ecr/ecr-command.sh install-ecr-public
```
abhay-krishna, line 26: nit: Erring on the side of pedantry, but I think this will work just fine without `bash` here. For example, https://github.com/aws/eks-distro/blob/main/Makefile#L88
```suggestion
	development/ecr/ecr-command.sh install-ecr-public
```

TerryHowe: Just following the style, but yeh, calling bash there would just result in two bashes

jaxesn (approved):
/lgtm

PR #176, opened 2021-02-11
We should link to all the community installs and let the customer decide

TerryHowe:
For some reason the link to the community installation page was removed and replaced with one method. This brings it back. Installing with `kubeadm` should probably be a last resort instead of our go to method


jaxesn (approved):
/lgtm

PR #179, opened 2021-02-16
Build local-images for aws iam and make shasums idempotent

TerryHowe:
Run checksums at end so we get all of them.

### DIFF
--- a/projects/kubernetes/kubernetes/build/create_release_checksums.sh
+++ b/projects/kubernetes/kubernetes/build/create_release_checksums.sh
@@ -42,7 +42,7 @@ rm -f $SHA256SUM
 rm -f $SHA512SUM
 echo "Writing artifact hashes to SHA256SUM/SHA512SUM files..."
 cd $ASSET_ROOT
-for file in $(find ${ASSET_ROOT} -type f ); do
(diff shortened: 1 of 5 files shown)


projects/kubernetes/kubernetes/build/create_release_checksums.sh
```diff
@@ -42,7 +42,7 @@ rm -f $SHA256SUM
 rm -f $SHA512SUM
 echo "Writing artifact hashes to SHA256SUM/SHA512SUM files..."
 cd $ASSET_ROOT
-for file in $(find ${ASSET_ROOT} -type f ); do
+for file in $(find ${ASSET_ROOT} -type f | grep -v '\.sha[25][51][62]' ); do
```
abhay-krishna, line 45:
```suggestion
for file in $(find ${ASSET_ROOT} -type f -not -path '*\.sha[25][51][62]' ); do
```
nit: Just avoiding pipes and another exit code to track. Thinking if we could use this
```bash
(~/test_find) touch test.txt
(~/test_find) touch test.py
(~/test_find) touch test.sha256
(~/test_find) touch test.sha512                                                                                                                                                                                                    
(~/test_find) find ./ -type f -not -path '*\.sha[25][51][62]'
./test.py
./test.txt
```

bnrjee (approved):

PR #188, opened 2021-02-17
on make test failures check for top flakes

jaxesn:
There are a number of flakes in upstream kubernetes.  There is a pretty good doc [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/flaky-tests.md) on it which links to the various dashboards they use to track the flakes.  

Bumping our retries to 5 and adding a top_flakes file, which I pulled from the latest [upstream](http://storage.googleapis.com/k8s-metrics/flakes-latest.json).  If tests are still failing after 5 tries, compare the failures to the top_flakes doc and if there are no failures which aren't listed, pass the tests.  

When running the presubmit jobs going forward, if we notice what appear to be flakes, we should validate with the upstream list and add to our top_flakes file.

### DIFF
--- a/projects/kubernetes/kubernetes/build/run_tests.sh
+++ b/projects/kubernetes/kubernetes/build/run_tests.sh
@@ -13,8 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
-set -x
 set -o errexit
 set -o nounset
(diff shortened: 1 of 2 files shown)


TerryHowe (approved):
/lgtm

PR #194, opened 2021-02-18
Generate tags based on release branch and release

TerryHowe:


development/kops/validate_eks.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
bnrjee, line 1: We should call this script.

TerryHowe: Done

bnrjee (approved):

PR #195, opened 2021-02-18
Fix go runner tagging

TerryHowe:


projects/kubernetes/kubernetes/Makefile
```diff
@@ -23,12 +23,8 @@ KUBE_BASE_TAG?=${KUBE_BASE_GIT_TAG}-eks-${RELEASE_BRANCH}-${RELEASE}
 GO_RUNNER_IMAGE?=$(KUBE_BASE_REPO)/$(GO_RUNNER_IMAGE_NAME):$(KUBE_BASE_TAG)
 KUBE_PROXY_BASE_IMAGE?=$(KUBE_BASE_REPO)/$(KUBE_PROXY_BASE_IMAGE_NAME):$(KUBE_BASE_TAG)
 
-# Since the image won't be available for the local-images target,
-# need to override based on an image that is already available.
-# We want to a better way than this to avoid having to change often.
-PRESUBMIT_KUBE_BASE_TAG?=v0.4.2-ea45689a0da457711b15fa1245338cd0b636ad4b
-PRESUBMIT_GO_RUNNER_IMAGE?=$(KUBE_BASE_REPO)/$(GO_RUNNER_IMAGE_NAME):$(PRESUBMIT_KUBE_BASE_TAG)
-PRESUBMIT_KUBE_PROXY_BASE_IMAGE?=$(KUBE_BASE_REPO)/$(KUBE_PROXY_BASE_IMAGE_NAME):$(PRESUBMIT_KUBE_BASE_TAG)
+PRESUBMIT_GO_RUNNER_IMAGE?=$(KUBE_BASE_REPO)/$(GO_RUNNER_IMAGE_NAME):$(KUBE_BASE_TAG)
```
abhay-krishna, line 26: We don't have an image with tag v0.4.2-eks-1-18-1 in our private ECR repo right? Should we revert to hardcoding it to a particular tag which is available?

TerryHowe: I think we need to either get that artifact in there or use build public ecr. If anything merged that built kubernetes/release or everything for 18, the presubmit for this would probably pass.

abhay-krishna: We probably need to get the main postsubmit kicked off with just a change in the kubernetes-release folder.

TerryHowe:
We no longer have a job that builds to the private ecr repo, so https://github.com/aws/eks-distro-prow-jobs/pull/120 will change the presubmit so we use the repo we are building to

bnrjee (approved):

PR #200, opened 2021-02-18
Fix checksum path

TerryHowe:
This isn't terribly important, but we generate our sha sum files with `./` in the file names and this change gets rid of that. The kubernetes shasums do not have this, this covers etcd, cni, and coredns.


jaxesn (approved):
/lgtm

PR #205, opened 2021-02-19
Copy up release channel and CRDs

TerryHowe:
This artifacts need to be pushed for 1-19


release/crds/releasechannels.distro.eks.amazonaws.com-v1alpha1.yaml
```diff
@@ -0,0 +1,75 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
```
vivek-koppuru, line 4: Shouldn't CRDs be defined in our build-tooling repo like here or are we moving them here instead? https://github.com/aws/eks-distro-build-tooling/tree/main/release/config/crds

release/1-18/1-18.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: distro.eks.amazonaws.com/v1alpha1
```
vivek-koppuru, line 1: Let's add the Apache License configuration to these files as well

TerryHowe: While # is a valid comment in yaml, I'm not a huge fan of uploading these artifacts with comments in them, nor am I a big fan of some logic to strip the comments off.

release/crds/releasechannels.distro.eks.amazonaws.com-v1alpha1.yaml
TerryHowe: Having two repos is a constant source of extra work and the risk of more moving parts, which is why I put them here.

release/1-18/1-18.yaml
vivek-koppuru: Why do we need to strip comments? Also it's a requirement to add the Apache License to our source code I thought.

TerryHowe: It just seems really odd to have that copyright comment in the middle of an artifact. We can just skip automating this and someone can copy these files manually with no copyright.

release/crds/releasechannels.distro.eks.amazonaws.com-v1alpha1.yaml
TerryHowe: Okay, I changed this to pull the CRDs out of the other repo

release/1-18/1-18.yaml
TerryHowe: I changed this to grep out the comments before uploading them.

kschumy (approved):
/lgtm

PR #210, opened 2021-02-19
Run a build of 1.18-1 so we have all artifacts

TerryHowe:
I made a change to the k8s/k8s/Makefile and the presubmit fails because go-runner is not build in the build account private repo. This should build the needed artifacts.


jaxesn (approved):
/lgtm

PR #213, opened 2021-02-21
Fix the preflight check

TerryHowe:
If preflight fails it gets set to "false" which doesn't work with this condition.


jaxesn (approved):
/lgtm

abhay-krishna (approved):
/lgtm
/approve

PR #226, opened 2021-03-02
Use one build badge

TerryHowe:
Just postsubmit badges. Prow does not allow you to set the text in the badge, so I put them in a table:

https://github.com/TerryHowe/eks-distro/blob/badges/README.md


abhay-krishna (approved):
/lgtm

PR #242, opened 2021-03-18
add attribution generation for containernetworking

jaxesn:
This introduces attribution generation into eks-d.  We have previously generated these files once and never updated them.  This PR only adds it to containernetworking to show the pattern.  I have another PR ready with all the others changed and based on the feedback here I will update and send it after this merges.

This PR is broken into two commits. 1)adding the handling to container networking and 2) adding the make target to support the periodic [job](https://github.com/aws/eks-distro-prow-jobs/pull/136) updating attribution files on a daily basis, if there are changes.

There are two cases where the attribution file needs to change. 1) the git_tag file changes because we are updating to a new version therefore deps could and probably did change 2) the golang version changes in the builder-base since this version does show up in the attribution file, by design and requirement.

Because the go version matters and the generation script requires some setup, we decided it felt best to add a periodic job to update these instead of requiring developers to do it manually.  

The attribution file for containernetworking actually does need updating, but I am not checking it in in this PR instead waiting to test the periodic. 


build/lib/common.sh
```diff
@@ -67,7 +67,71 @@ function build::common::generate_shasum() {
   cd -
 }
 
-# TODO: replace with go licenses tool which was used to create ATTRIBUTION.txt files
+
+function build::gather_licenses_new() {
```
jaxesn, line 71: Doing this change in multiple PRs so introducing temporary _new function.  In my last PR changing the license handling across all projects, I will rename this and remove the old function.

build/lib/common.sh
```diff
@@ -78,7 +142,16 @@ function build::gather_licenses() {
      -o -name 'LICENSE.code' -o -name 'LICENCE.md' \) | while IFS= read -r NAME; do mkdir -p "${outputdir}/$(dirname $NAME)" && cp  "$NAME" "${outputdir}/${NAME}"; done
 }
 
-function build::common::use_go_version() {
+function build::common::remove_go_path() {
+  # This is the path where the specific go binary versions reside in our builder-base image
+  export PATH=$(echo "$PATH" | sed -e "s/^\/go\/go[0-9\.]*\/bin://")
```
TerryHowe, line 147: This looks troublesome, can't we just put the version we want first? This won't work in my env, although I'm not sure that matters.

TerryHowe (approved):
/lgtm

build/lib/common.sh
jaxesn: yeah i was thinking about that too, we probably can. ill make another rev on this.

PR #243, opened 2021-03-18
Download 1.20 kops

TerryHowe:


development/kops/eks-d.tpl
```diff
@@ -66,6 +66,7 @@ spec:
   # Metrics Server will be supported with kops 1.19
   metricsServer:
     enabled: true
+    insecure: true
```
TerryHowe, line 69: The hold is to make sure this change doesn't break 1.18 and 1.19

TerryHowe:
This is good to go

PR #246, opened 2021-03-22
remove volume plugin patch

jyotimahapatra:
*Description of changes:*
Based on https://kubernetes.io/docs/setup/release/notes/  

```
Kube-controller-manager: volume plugins can be restricted from contacting local and loopback addresses by setting --volume-host-allow-local-loopback=false, or from contacting specific CIDR ranges by setting --volume-host-cidr-denylist (for example, --volume-host-cidr-denylist=127.0.0.1/28,feed::/16) (#91785, @mattcary) [SIG API Machinery, Apps, Auth, CLI, Network, Node, Storage and Testing]
```

We don't need the eks patch anymore, since the fix is available in upstream.


jyotimahapatra:
@kschumy @micahhausler @TerryHowe PTAL

TerryHowe (approved):
/lgtm

PR #257, opened 2021-03-25
add attribution generation for remaining projects

jaxesn:
Adding attribution generation to the remaining projects.

For projects with release branch folders, the attribution files have been moved into those folders. I did not update any attribution files, just copied from some from what was already there.  Once this is merged, the bot will create a PR updating all files.


abhay-krishna (approved):
/lgtm

jaxesn:
On hold until releases are done, dont want to cause any issues there.

projects/kubernetes/kubernetes/build/lib/binaries.sh
```diff
@@ -13,14 +13,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-function build::binaries::use_go_version_k8s() {
+function build::binaries::get_go_version_k8s() {
     local -r releasebranch="$1"
 
     if [[ "$releasebranch" == "1-18" ]]; then
-        build::common::use_go_version "1.13"
```
TerryHowe, line 20: Can we maybe put a GOLANG_VERSION file in the various release branch directories instead?

jaxesn: I like that idea.  Kubernetes is the only project so far using different go versions per release branch and this was the pattern put in place for that.  I can open a follow up that moves it to a GOLANG_VERSION.

projects/coredns/coredns/Makefile
```diff
@@ -3,6 +3,7 @@ RELEASE_BRANCH?=1-18
 RELEASE?=$(shell cat $(BASE_DIRECTORY)/release/$(RELEASE_BRANCH)/RELEASE)
 ARTIFACT_BUCKET?=my-s3-bucket
 GIT_TAG?=$(shell cat $(RELEASE_BRANCH)/GIT_TAG)
+GOLANG_VERSION?=1.13
```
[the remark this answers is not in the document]
jaxesn: I believe when coredns was updated it was checked what version they were building with so 1.13 was left, but yeah whenever people update a GIT_TAG file we should make sure they are validating the upstream version of go that is being used.

TerryHowe (approved):
/lgtm

PR #264, opened 2021-03-27
fix retry logic for kubernetes unit tests

jaxesn:
When the kubernetes unit tests would fail 5 times, if the only failures are known flakes it is supposed to still pass.  There was a bug in the way it checked for failures in that in prow the ARTIFACTS env var is set to /logs/artifacts but the script was assuming ./_artifacts.


TerryHowe (approved):
/lgtm

PR #267, opened 2021-03-28
fix unit test rety logic by adding gotpath to path for gotestsum

jaxesn:
gotestsum was being installed but was not on the PATH


jaxesn:
I forced a test to failure and tested to make sure this workflow went all the way through in prow.

TerryHowe (approved):
/lgtm

PR #268, opened 2021-03-29
use same version of kops for all release branches

jaxesn:
Looking at the kops release compat [page](https://kops.sigs.k8s.io/welcome/releases/) I wonder if we should just use the latest kops version for all k8s versions?  When updating to a new k8s version, we would then just bump the kops version to the respective version for that new k8s version.  Based on the historical release schedule, this may mean we end up typically using beta releases of kops, but that is what we have been using so far.

### DIFF
--- a/development/kops/cluster_wait.sh
+++ b/development/kops/cluster_wait.sh
@@ -23,7 +23,7 @@ $PREFLIGHT_CHECK_PASSED || exit 1
 # Add IAM configmap
 COUNT=0
 echo 'Waiting for cluster to come up...'
-while ! kubectl apply -f ./${KOPS_CLUSTER_NAME}/aws-iam-authenticator.yaml
+while ! kubectl --context $KOPS_CLUSTER_NAME apply -f ./${KOPS_CLUSTER_NAME}/aws-iam-authenticator.yaml
 do
     sleep 5
     COUNT=$(expr $COUNT + 1)
--- a/development/kops/eks-d.tpl
+++ b/development/kops/eks-d.tpl
@@ -63,7 +63,6 @@ spec:
     image: {{ .kube_scheduler.repository }}:{{ .kube_scheduler.tag }}
   kubeProxy:
     image: {{ .kube_proxy.repository }}:{{ .kube_proxy.tag }}
-  # Metrics Server will be supported with kops 1.19
   metricsServer:
     enabled: true
     insecure: true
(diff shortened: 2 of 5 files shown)


development/kops/cluster_wait.sh
```diff
@@ -23,7 +23,7 @@ $PREFLIGHT_CHECK_PASSED || exit 1
 # Add IAM configmap
 COUNT=0
 echo 'Waiting for cluster to come up...'
-while ! kubectl apply -f ./${KOPS_CLUSTER_NAME}/aws-iam-authenticator.yaml
+while ! kubectl --context $KOPS_CLUSTER_NAME apply -f ./${KOPS_CLUSTER_NAME}/aws-iam-authenticator.yaml
```
jaxesn, line 26: If ever running locally and potentially modifying the kubectl context or creating more than one cluster at a time, this ensures we are operating on the correct cluster.

development/kops/eks-d.tpl
```diff
@@ -74,30 +73,14 @@ spec:
   kubeDNS:
     provider: CoreDNS
     coreDNSImage: {{ .coredns.repository }}:{{ .coredns.tag }}
-    externalCoreFile: |
```
jaxesn, line 77: this was needed when we were using kops 1.18 and coredns 1.7. This is now the default in kops 1.19+

development/kops/eks-d.tpl
```diff
@@ -74,30 +73,14 @@ spec:
...
-          loadbalance
-          reload
-      }
   masterKubelet:
     podInfraContainerImage: {{ .pause.repository }}:{{ .pause.tag }}
-  # kubelet might already be defined, append the following config
   kubelet:
     podInfraContainerImage: {{ .pause.repository }}:{{ .pause.tag }}
     anonymousAuth: false
+    # for 1.19 and above webhook auth is the default mode
```
jaxesn, line 81: for metrics server to work these need to be set.

development/kops/eks-d.tpl
TerryHowe: I'm not sure where that came from

development/kops/install_requirements.sh
```diff
@@ -31,18 +31,7 @@ fi
...
-    else
-        if [ "${RELEASE_BRANCH}" == "1-20" ]
-        then
-            KOPS_VERSION="v1.20.0-beta.2"
-        else
-            KOPS_VERSION="v1.19.0-beta.3"
-        fi
-    fi
-
+    KOPS_VERSION="v1.20.0-beta.2"
```
TerryHowe, line 34: I was thinking someday it might be nice to have a KOPS_VERSION file like we have GIT_TAG maybe.

TerryHowe (approved):
/lgtm

PR #269, opened 2021-03-29
introduce golang-version file for kubernetes release branches

jaxesn:
After adding the golang version to the makefiles to support attribution generation, @TerryHowe and I were thinking it may be a better pattern to pull golang version from a file per release branch instead of using the bash helper.


TerryHowe (approved):
/lgtm

TerryHowe:
Sweet thanks!

PR #279, opened 2021-03-31
Get rid of DEVELOPMENT

TerryHowe:
This DEVELOPMENT variable was used in a strange way and we should just stick with the new RELEASE_ENVIRONMENT. This changes the development build so you use the real base image consistent with how prow does it. Our instructions already say to use the base image.


jaxesn (commented):
I think the issue here is that the builder base isnt a public image so we had the development stuff in place to make building locally easier and in theory allow others outside of aws to build locally.

TerryHowe:
Yeh, but we document building the base as the first step https://distro.eks.amazonaws.com/users/build/ 

Maybe the solution would be to make the base image public though. Using that old amazonlinux image isn't great.

abhay-krishna (approved):
/lgtm

PR #280, opened 2021-03-31
Exit PR creation script gracefully if no files were changed

abhay-krishna:
We don't want the script to invoke `git commit` and fail if no files were modified, i.e., thus no files were added by `git add`.


jaxesn (approved):
/lgtm

PR #287, opened 2021-04-05
use upstream log-dump script to gather logs during e2e test

jaxesn:
The kubernetes/test-infra repo had a log collection script they use in the various prow jobs to test kubernetes clusters.  It handles a lot more than we really need, windows node/gke/etc, but I was thinking it would be easier to use it straight up than take on the work off modifying for our use case and maintaining.  I have left the license intact and added a short readme explaining what it is.

### DIFF
--- a/development/kops/gather_logs.sh
+++ b/development/kops/gather_logs.sh
@@ -19,14 +19,6 @@ BASEDIR=$(dirname "$0")
 source ${BASEDIR}/set_environment.sh
 $PREFLIGHT_CHECK_PASSED || exit 1
 
-UPSTREAM_PATH=$BASE_DIRECTORY/../k8s.io/test-infra/logexporter/cluster/log-dump.sh
-if [ ! -f "$UPSTREAM_PATH" ]; then
-    echo "k8s.io/test-infra is missing, skipping log gathering"
-    exit
-fi
-
-BASE_DIRECTORY=$(git rev-parse --show-toplevel)
-
 export LOG_DUMP_SAVE_SERVICES="kops-configuration"
 export KUBERNETES_PROVIDER="none" # the script from upstream doesnt seem to handle aws exactly right
 export LOG_DUMP_SSH_KEY="~/.ssh/id_rsa"
@@ -62,4 +54,4 @@ function scp() {
 }
 export -f scp
 
-source $UPSTREAM_PATH $ARTIFACTS
+source ./log-dump/log-dump.sh $ARTIFACTS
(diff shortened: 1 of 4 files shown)


bnrjee (dismissed):
We need to remove/replace each occurrence of the word master.

development/kops/log-dump/log-dump.sh
```diff
@@ -0,0 +1,917 @@
...
+  fi
+}
+
+function log-dump-ssh() {
+  local host="$1"
+  local cmd="$2"
+
+  if [[ "${gcloud_supported_providers}" =~ ${KUBERNETES_PROVIDER} ]]; then
+    for (( i=0; i<5; i++)); do
+      if gcloud compute ssh --ssh-flag="-o LogLevel=quiet" --ssh-flag="-o ConnectTimeout=30" --project "${PROJECT}" --zone="${ZONE}" "${host}" --command "echo test > /dev/null"; then
```
bnrjee, line 265: Do we want to use this?

development/kops/gather_logs.sh
```diff
@@ -0,0 +1,57 @@
...
+    else
+        filter="Name=tag:k8s.io/role/node,Values=1"
+    fi
+    nodes=$(aws ec2 describe-instances \
+        --filters "Name=tag:KubernetesCluster,Values=${KOPS_CLUSTER_NAME}" "$filter" \
+        --query "Reservations[*].Instances[*].PublicDnsName" --output text)
+
+    # contaner logs are symlinks and need a slightly differnet chmod
+    # log-dump-ssh comes the upstream log-dump.sh script
+    for node in $nodes ; do log-dump-ssh $node "sudo chmod a+r /var/log/containers/*" || true ; done
```
bnrjee, line 44: We may want to store them in /logs/artifacts.

jaxesn: On prow they should be because the artifacts env is set

jaxesn:
Fixed the non-inclusive language and added a note to the readme if we ever need to update from upstream and reproduce the changes.

development/kops/gather_logs.sh
TerryHowe: Up on line 28 I assume

TerryHowe (approved):
/lgtm

development/kops/gather_logs.sh
jaxesn: Right, this is just the path on the instances created by kops. The log-dump script copies all of those into the artifacts path in the prow job pod.

PR #291, opened 2021-04-07
Only update files that do not exist

TerryHowe:

### DIFF
--- a/release/s3_sync.sh
+++ b/release/s3_sync.sh
@@ -44,6 +44,7 @@ cd ${PREFIX_DIR}
 for CRD
 in *yaml
 do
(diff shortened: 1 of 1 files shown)


release/s3_sync.sh
```diff
@@ -44,6 +44,7 @@ cd ${PREFIX_DIR}
 for CRD
 in *yaml
 do
+  aws s3 ls s3://${ARTIFACT_BUCKET}/${PREFIX_DIR}/${CRD} ||
```
jaxesn, line 47: The way this is spaced out this looks like 2 different commands. Could we make this one line or add a `\` or something to make it more clear its one command?

TerryHowe: Done

jaxesn (approved):
/lgtm

PR #292, opened 2021-04-08
Use release environment in kops

TerryHowe:


development/kops/set_environment.sh
```diff
@@ -16,7 +16,8 @@
 BASE_DIRECTORY=$(git rev-parse --show-toplevel)
 export DEFAULT_RELEASE_BRANCH=$(cat ${BASE_DIRECTORY}/release/DEFAULT_RELEASE_BRANCH)
 export RELEASE_BRANCH=${RELEASE_BRANCH:-"${DEFAULT_RELEASE_BRANCH}"}
-export DEFAULT_RELEASE=$(cat ${BASE_DIRECTORY}/release/${RELEASE_BRANCH}/RELEASE)
+RELEASE_ENVIRONMENT=${RELEASE_ENVIRONMENT:-development}
```
jaxesn, line 19: Could either way on this, but do we want to default to production instead so that if someone just creates a cluster using these scripts it will use the prod release?  and then in the build set it to development?  dont have a strong opinion either way.

TerryHowe: All the make files default to develop

jaxesn (approved):
/lgtm

PR #294, opened 2021-04-11
WIP - Add OCI image tarballs for etcd and coredns

abhay-krishna:
We need OCI image tarballs of etcd and coredns projects for building AMI's with EKS-D artifacts.

### DIFF
--- a/projects/coredns/coredns/Makefile
+++ b/projects/coredns/coredns/Makefile
@@ -34,6 +34,10 @@ endif
 binaries:
 	build/create_binaries.sh $(CLONE_URL) $(REPO) $(GIT_TAG) $(GOLANG_VERSION)
 
+.PHONY: image-tars
+image-tars:
+	build/create_image_tars.sh $(BASE_IMAGE) $(RELEASE_BRANCH) $(IMAGE)
+
 .PHONY: local-images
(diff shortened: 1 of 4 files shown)


projects/coredns/coredns/Makefile
```diff
@@ -57,6 +58,14 @@ images: binaries
 		--local dockerfile=./docker/linux \
 		--local context=. \
 		--output type=image,oci-mediatypes=true,name=$(IMAGE),push=true
+	buildctl \
+		build \
+		--frontend dockerfile.v0 \
+		--opt platform=linux/amd64 \
```
jaxesn, line 64: This is only building the amd64 whereas the one above is also building arm64. Is this is intentional?

abhay-krishna: /hold
Good call. This PR is actually a work in progress. Wanted to make a few more changes like adding checksums and placing artifacts in the appropriate directory since we will probably be releasing this. I'll add ARM64 builds along with that change.

abhay-krishna:
Discussed with Vignesh, we probably do not need this as we can pass in the actual image endpoint to CAPI when bootstrapping, instead of consuming the image TAR in the image-builder phase.

PR #300, opened 2021-04-14
kubernetes do not call local-images in release target

jaxesn:
We added the local-images call to the release target to support the kops testing because the oci tarballs are needed.  However, due to the fact that the local-images is intended to be used only by the build target and not the release target, a hardcoded "hack" was added to make local-images work when adding new versions of kubernetes.  This means that our oci images produced by the release targets are not 100% correct since they are using the hardcoded go-runner tag.  This also makes spinning up new dev stacks/envs difficult because we assume that hardcoded go-runner image exist (I am going to look at this in a future pr).

This PR changes the kubernetes makefile to instead of calling the local-images target, the images target calls the helper create-images script two times, one of local oci tarballs and one for pushing to the registry.

### DIFF
--- a/projects/kubernetes/kubernetes/build/lib/images.sh
+++ b/projects/kubernetes/kubernetes/build/lib/images.sh
@@ -141,8 +141,12 @@ function build::images::pause_tar(){
     local -r image_tag="$3"
     local -r context_dir="$4"
     local -r bin_dir="$5"
+    local -r skip_arm="$6"
 
     for platform in "${KUBE_LINUX_IMAGE_PLATFORMS[@]}"; do
+        if [ "$platform" == "arm64" ] && [ $skip_arm == true ]; then
+            continue
+        fi
         buildctl \
             build \
             --frontend dockerfile.v0 \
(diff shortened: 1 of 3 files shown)


projects/kubernetes/kubernetes/build/lib/images.sh
```diff
@@ -133,3 +133,48 @@ function build::images::docker::push(){
...
+function build::images::pause_tar(){
+    local -r go_runner_image="$1"
+    local -r version="$2"
+    local -r image_tag="$3"
+    local -r context_dir="$4"
+    local -r bin_dir="$5"
+    local -r skip_arm="$6"
+
+    for platform in "${KUBE_LINUX_IMAGE_PLATFORMS[@]}"; do
+        if [ "$platform" == "arm64" ] && [ $skip_arm == true ]; then
```
jaxesn, line 147: arm builds for pause fail on the presubmit cluster, i believe because we dont have qemu setup fully on the fargate nodes.  Skipping arm in this case is what we did before, but now we build both amd64 and arm64 in post submit.

TerryHowe (approved):
/lgtm

PR #301, opened 2021-04-14
Make kops domain name configurable

abhinavmpandey08:
kops domain name is currently set to a static value `prod-build-pdx.kops-ci.model-rocket.aws.dev`
This PR parameterizes that kops domain so a custom one can be passed in without having to modify the `prow.sh` script. 


TerryHowe (approved):
/lgtm

PR #319, opened 2021-04-19
Use public ECR eks-distro-base as base image

abhay-krishna:
Now that we have eks-distro-base image in public ECR, we can use them as base images for building EKS-D.


projects/coredns/coredns/Makefile
```diff
@@ -15,7 +15,7 @@ AWS_REGION?=us-west-2
 AWS_ACCOUNT_ID?=$(shell aws sts get-caller-identity --query Account --output text)
 
 MAKE_ROOT=$(shell cd "$(shell dirname "${BASH_SOURCE[0]}")" && pwd -P)
-BASE_REPO?=$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com
+BASE_REPO?="public.ecr.aws/h1r8a7l5"
```
TerryHowe, line 18: this is our development build. maybe we should register a name for it so people know what they are getting. Another alternative would be move the base image build into this repository which would solve a lot of problems

EKS_DISTRO_BASE_TAG_FILE
```diff
@@ -1 +1 @@
-2021-03-22-1616436036
```
ttashi, line 1: is this meant for the release tag ? is this generated manually... any plan on build-system doing it ?

abhay-krishna: Yes this was a tagging scheme with date and Unix epoch, that we came up with so that we can identify when each new image was built and the ordering. This is especially since there is a periodic Prowjob that may also update this base image. The other image builds use commit hash Prow environment variables generated from PR's and merges , but the periodic does not have that, so to keep the tagging consistent within this image build, we chose this.

vignesh-goutham: So the plan is to have an automation that will update these, when a new image is built. Automated jobs would cut a PR and some human would approve it and merge it. Since these are private repos now, we are punting the automation tasks for now until we will be public.

The number of automation stuff is starting to grow, I'm going to start an umbrella issue that would track all the automation we have to do once we go public.

projects/coredns/coredns/Makefile
abhay-krishna: +1 on registering a name, but can we register a name on an unlisted repo? Also do you think we can push this image to the verified public ECR repo and consume from there, instead of using dev build?

EKS_DISTRO_BASE_TAG_FILE
abhay-krishna: +1
I wanted to do some testing on generating automated PR's for private repos too. It's just a matter of cloning them properly and seeing if Github CLI will support the operations on private repos.

ttashi: Thanks guys for clarifying.

PR #328, opened 2021-04-24
Added directions for component versions

kschumy:
*Issue #, if available:*

*Description of changes:*


docs/contents/runbook.md
```diff
@@ -32,6 +32,7 @@ All other changes are to this package.
    * docs/contents/releases/${RELEASE_BRANCH}/${RELEASE}/CHANGELOG
    * docs/contents/index.md
    * README
+   * development/pull-all.sh (make sure component versions are correct)
```
TerryHowe, line 35: That pull-all should be full automated so we don't need this step, but f9ine for now

kschumy: It was my impression that this would not be included in the automated release workflow, as it likely would only be done with the minor releases.

TerryHowe: What I meant by that comment is if it just got the GIT_TAG values for all the various images, it wouldn't need to be a step here at all.

kschumy:
We're changing pull-all.sh, so this PR isn't needed

PR #331, opened 2021-04-29
Rerun sonobuoy failed test

bnrjee:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/development/kops/run_sonobuoy.sh
+++ b/development/kops/run_sonobuoy.sh
@@ -44,6 +44,8 @@ do
 done
 
 results=$(./sonobuoy --context ${KOPS_CLUSTER_NAME} retrieve)
+./sonobuoy e2e --wait $results --rerun-failed --kube-conformance-image k8s.gcr.io/conformance:v1.19.6
+results=$(./sonobuoy --context ${KOPS_CLUSTER_NAME} retrieve)
 mv $results "./${KOPS_CLUSTER_NAME}/$results"
 results="./${KOPS_CLUSTER_NAME}/$results"
 mkdir ./${KOPS_CLUSTER_NAME}/results


TerryHowe (approved):
/lgtm

development/kops/run_sonobuoy.sh
```diff
@@ -43,6 +43,8 @@ do
   echo 'Waiting for the cluster to be ready...'
 done
 
+results=$(./sonobuoy --context ${KOPS_CLUSTER_NAME} retrieve)
+./sonobuoy e2e --wait $results --rerun-failed --kube-conformance-image k8s.gcr.io/conformance:v1.19.6
```
kschumy, line 47: Are we going to need to change this for 1.19.8?

jaxesn: We should check the result and loop similar to what we do above up to like 3 times retrying. And yeah remove this hardcoded image, use the var from above.

jaxesn (changes requested):

jaxesn:
I think there is a pretty easy way to test this workflow.  When you create the kops cluster, skip the set_nodeport_access script, which will cause a few of the tests to fail. To make it quick, you can pass the one test to the sonobuoy call so it only run one test which will fail, making it quick and easy to validate this. I think I can find the name of one of the tests that will fail when nodeport access isn't set in the morning and post it here for reference.

jaxesn:
This should be one of the test that will fail if we dont run that nodeport access script.  You can run it like this:

` sonobuoy run --e2e-focus "Services should be able to change the type from NodePort to ExternalName" ....`

development/kops/run_sonobuoy.sh
TerryHowe, line 47:
```suggestion
./sonobuoy e2e --wait $results --rerun-failed --kube-conformance-image ${CONFORMANCE_IMAGE}
```

TerryHowe:
This is almost 2 months old, I assume we don't need this. Reopen if we have something here.

PR #333, opened 2021-04-29
Consume latest eks-distro-base image from verified build-tooling public registry

abhay-krishna:
New eks-distro-base images will be pushed to the [build-tooling public ECR registry](https://gallery.ecr.aws/eks-distro-build-tooling/eks-distro-base) and we should consume it from there.


jaxesn (approved):
/lgtm

PR #334, opened 2021-04-29
Removed patch, incremented Kubernetes version for 1.18, and dev RELEASE

kschumy:
*Issue #, if available:*

*Description of changes:*


kschumy:
The failed test is both unexpected but not terribly surprising. I ran `make binaries` just fine locally, but I had noticed that the deleted patch and the subsequent one (the one now called "0017...") both dealt with the same file, which seems to be the one having the issue.

I'm looking into this now

kschumy:
Just realized that it built fine locally because I was building the default version, which is 1.19 🤦🏻‍♀️🤦🏻‍♀️🤦🏻‍♀️

kschumy:
/hold

The patches problem is more complicated than I originally thought. I'll unhold when I've resolved this

kschumy:
Kinda a hot mess situation with the patches. Closing and breaking into two prs

PR #337, opened 2021-04-30
overwrite clusterrole for coredns to have endpointslices perms

jaxesn:
If we do want to ship 1.8.3 this should fix the conformance test runs.

I got this from the PR in the codedns repo https://github.com/coredns/deployment/pull/246

Interestingly, upstream kubernetes only add 1.8.0 in the 1.21 branch and when they added it they did not seem to update the clusterrole, not sure if this is something they havent tested yet? https://github.com/kubernetes/kubernetes/pull/99752/files

I want to make sure 1.8.3 is what the EKS team wants us to include in the 1.20 release.

### DIFF
--- a/development/kops/cluster_wait.sh
+++ b/development/kops/cluster_wait.sh
@@ -35,5 +35,23 @@ do
     echo 'Waiting for cluster to come up...'
 done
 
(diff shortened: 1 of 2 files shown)


TerryHowe (approved):
/lgtm

PR #340, opened 2021-05-01
fix dev + prod release script to use new public base

jaxesn:
We missed these scripts when updating for the new public image.


abhay-krishna (approved):
/lgtm

PR #346, opened 2021-05-05
Add source epoch notes

TerryHowe:


jaxesn:
I think the easier way is to run `make update-version` in the kubernetes project, that should recreate that git file.

kschumy:
> I think the easier way is to run `make update-version` in the kubernetes project, that should recreate that git file.

...dang it.

PR #347, opened 2021-05-05
fix ecr-public login to check for running docker vs if it is installed

jaxesn:
We added docker to the builder-base image so now the wrong block is being run.  Changed condition to check for docker and make sure its running.

We might be able to remove a lot of this logic since the ecr-helper now supports ecr-pubic, but we can look into that a future date.


TerryHowe (approved):
/lgtm

PR #361, opened 2021-05-18
Removed CRD from all index.md files

kschumy:
*Issue #, if available:*

*Description of changes:*
Some of the shasums were wrong. We decided it'd just be better to remove the CRD yaml and point out more clearly where people can get this information


TerryHowe:
I like this change, but what ever happened to that ticket to generate this info in a more readable format?

kschumy:
> I like this change, but what ever happened to that ticket to generate this info in a more readable format?

@TerryHowe Do you want it be like this?

| Name | Version | URI |
|------|---------|-----|
| coredns | v1.8.3 |  public.ecr.aws/eks-distro/coredns/coredns:v1.8.3-eks-1-20-1 |
| etc. | etc. | etc. |

If not, what did you have in mind?

TerryHowe:
> > I like this change, but what ever happened to that ticket to generate this info in a more readable format?
> 
> @TerryHowe Do you want it be like this?
> 
> Name	Version	URI
> coredns	v1.8.3	public.ecr.aws/eks-distro/coredns/coredns:v1.8.3-eks-1-20-1
> etc.	etc.	etc.
> If not, what did you have in mind?


I like that format. How about it @jaxesn

jaxesn:
Yeah that looks good to me.  Can we automate that generation?

kschumy:
Working on having this be generated so I don't have to fix all these manually, and we don't have to deal with this in the future either

kschumy:
This should be it.

TerryHowe (approved):
/lgtm

PR #365, opened 2021-05-20
Remove yum updates from Dockerfile since the base image has already run it

abhay-krishna:
Removing `yum update` from kube-proxy-base Dockerfile since it uses eks-distro-base as he base image and that already has been `yum update`d


jaxesn (approved):
/lgtm

PR #373, opened 2021-05-25
Added SSM agent install to kops

micahhausler:
*Issue #, if available:*
N/A

*Description of changes:*

Added SSM agent installation to the kops template. Amazon SSM is not required for EKS-D, but it is leveraged for installing internal agents in our CI processes.

### DIFF
--- a/development/kops/eks-d.tpl
+++ b/development/kops/eks-d.tpl
@@ -24,6 +24,13 @@ spec:
       name: a
     memoryRequest: 100Mi
     name: events
+  externalPolicies:
+    node:
+    - arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore
+    master:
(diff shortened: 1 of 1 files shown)


development/kops/eks-d.tpl
```diff
@@ -24,6 +24,13 @@ spec:
       name: a
     memoryRequest: 100Mi
     name: events
+  externalPolicies:
```
TerryHowe, line 27: Is this going to prevent customers from creating a kops cluster with these scripts. I thought we had testing specific stuff isolated out ouf the default kops creat scripts.

micahhausler: No, this should work fine. This policy is the minimum required for SSM agent to operate, it is _not_ administrative SSM permissions.

TerryHowe (approved):
/lgtm

PR #376, opened 2021-06-02
update kubernetes update instructions

jaxesn:

### DIFF
--- a/docs/contents/runbook.md
+++ b/docs/contents/runbook.md
@@ -6,24 +6,34 @@ unless there is a change.
 Prow jobs are in the [eks-distro-prow-jobs repo](https://github.com/aws/eks-distro-prow-jobs/tree/main/jobs/aws/eks-distro).  
 All other changes are to this package. 
 
-### Minor Release
-1. **Prow Build Job**: Create and merge Prow build job for new release (build-1-20-postsubmits.yaml)
-1. **Create EKS Distro PR**: Create various release components:
+### Create new Kubernetes minor version
+1. **Prow Build Job**: Create and merge Prow build job for new release
+   * copy previous job, ex: build-1-20-postsubmits.yaml, and change RELEASE_BRANCH
+1. **Prow Presubmit Job**: Create presubmit for new release
+   * copy previous job, ex: kubernetes-1-20-presubmits.yaml, and change RELEASE_BRANCH
+1. **Create EKS Distro PR**: Create kubernetes RELEASE_BRANCH:
    * release/${RELEASE_BRANCH}/development/RELEASE (set to 0)
    * release/${RELEASE_BRANCH}/production/RELEASE (set to 0)
-   * projects/kubernetes/kubernetes/${RELEASE_BRANCH}/GIT_TAG
-   * projects/kubernetes/kubernetes/${RELEASE_BRANCH}/KUBE_GIT_VERSION_FILE
-   * Create GIT_TAG files for other components (etcd, coredns, ...)
-   * Add support in kops script for new release
-1. **Prow Presubmit Job**: Create presubmit for new release (kubernetes-1-20-presubmits.yaml)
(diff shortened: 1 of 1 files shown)


docs/contents/runbook.md
```diff
@@ -6,24 +6,34 @@ unless there is a change.
...
 All other changes are to this package. 
 
-### Minor Release
-1. **Prow Build Job**: Create and merge Prow build job for new release (build-1-20-postsubmits.yaml)
-1. **Create EKS Distro PR**: Create various release components:
+### Create new Kubernetes minor version
+1. **Prow Build Job**: Create and merge Prow build job for new release
+   * copy previous job, ex: build-1-20-postsubmits.yaml, and change RELEASE_BRANCH
+1. **Prow Presubmit Job**: Create presubmit for new release
+   * copy previous job, ex: kubernetes-1-20-presubmits.yaml, and change RELEASE_BRANCH
```
kschumy, line 13: Should these be in two separate PRs?

jaxesn: they dont have to be.  I was debating on how to structure this.  Do you see each of these bullets as a new PR?  technically down below they need to make presubmit jobs for all the projects that are versioned.  Those could all be in this same PR, but they could also be separate.  I open to suggestions.

docs/contents/runbook.md
```diff
@@ -6,24 +6,34 @@ unless there is a change.
...
-   * projects/kubernetes/kubernetes/${RELEASE_BRANCH}/KUBE_GIT_VERSION_FILE
-   * Create GIT_TAG files for other components (etcd, coredns, ...)
-   * Add support in kops script for new release
-1. **Prow Presubmit Job**: Create presubmit for new release (kubernetes-1-20-presubmits.yaml)
-1. **Prow Release Jobs**: Create release jobs
+   * projects/kubernetes/kubernetes/${RELEASE_BRANCH}/GIT_TAG (set to upstream tag)
+   * copy ATTRIBUTION.txt and GOLANG_VERSION from previous release branch folder to the new
+   * Add release/${RELEASE_BRANCH}/patches with the patches from the EKS team
+   * Create projects/kubernetes/kubernetes/${RELEASE_BRANCH}/KUBE_GIT_VERSION_FILE using
+      * `RELEASE_BRANCH=${RELEASE_BRANCH} make update-version`
```
kschumy, line 21: Did you try running this? I *think* it might need the RELEASE file to exist. But I'm not positive and will double check

kschumy:
I like how "Create new Kubernetes minor version" and "Release" are split. This makes way more sense than how we had it before

docs/contents/runbook.md
kschumy: I interpret each bullet point as a PR, but idk if anyone else see it that way. It's fine if we keep them split and let people decide one or two PRs.

For the presubmit jobs on the projects, do you mean moving those changes up to here? I'd prefer those stay where they are because it'd be easier to remember which jobs you need to update when you're also updating the versions in this eks-distro

kschumy: Though I guess that means that step is two PRs (1 per repo), but again I think I might be alone in thinking bullet point == PR

jaxesn: I sorta read them that way to tbh.  I think the other projects prow jobs make sense where they are in the instructions.  I am cool to leave it this way and we can see how people follow and if its too confusing, we can tweak.

docs/contents/runbook.md
jaxesn: Yeah I ran it.  By release file, you mean the development/production release files we create above set to 0?

kschumy: Yeah I just now saw that re set it to 0 before this, so this should be good

kschumy (approved):
/lgtm

PR #399, opened 2021-06-21
update kops for 1.21 support

jaxesn:
We need to update to the 1.21 for kubernetes 1.21 support.  We typically end up using beta releases so that part is not a concern.  I have not tested this, but will watch the postsubmit test runs to make sure the previous versions still work.

### DIFF
--- a/development/kops/install_requirements.sh
+++ b/development/kops/install_requirements.sh
@@ -31,7 +31,7 @@ fi
 if [ ! -x ${KOPS} ]
 then
     echo "Determine kops version"
-    KOPS_VERSION="v1.20.0"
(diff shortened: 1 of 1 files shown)


abhay-krishna (approved):
/lgtm
I expect kOps to support previous 4 or 5 versions of k8s according to the [compatibility matrix](https://kops.sigs.k8s.io/welcome/releases/#compatibility-matrix). Also the [1.21 beta release notes](https://github.com/kubernetes/kops/releases/tag/v1.21.0-beta.3) state that they're dropping support only for k8s 1.13 and 1.14, so they don't apply to us.

PR #401, opened 2021-06-22
do not ignore delete cluster failure

jaxesn:
If delete fails we were still deleting the store which makes cleanup harder.  I think its probably ok to fail if delete fails?


TerryHowe (approved):
/lgtm

PR #403, opened 2021-06-22
kubernetes/release: introduce release branches

jaxesn:
We havent been staying in sync on the kubernetes/release project, which is where we get the iptables script for kube-proxy and the source for go-runner.  This has probably been ok because they do not change too often, but updating to match upstream and adding to the runbook.


TerryHowe (approved):

abhay-krishna (approved):
/lgtm

PR #404, opened 2021-06-22
Use release branch when fetching kubernetes release git tag

abhay-krishna:
Since the kubernetes/release project was separated based on release branches, we would need to access the git tags under those folders using the release git tags.


jaxesn (approved):
/lgtm

PR #388, opened 2021-06-12
Updated KUBE_GIT_VERSION_FILE and bumped prod RELEASE for 1.20-2

kschumy:
*Issue #, if available:*

*Description of changes:*
Hold until release

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #413, opened 2021-06-26
Bumped prod RELEASE for 1.20-2

kschumy:
*Issue #, if available:*

*Description of changes:*
Replaces #388, which combined these changes with other we decided to hold. 

Out of an abundance of caution since this change is so important (and also because of my usual crippling anxiety about messing things up), I decided to open an new PR instead of modifying that one in case there was something weird that would git-wise by rolling back that commit in that PR.


TerryHowe (approved):
/lgtm

PR #427, opened 2021-07-06
bumped kubernetes version to 1-21-2

ahreehong:
Bumped kubernetes version to 1-21-2 for the 1-21 minor release.


TerryHowe (commented):
I don't think the runbook is exactly right and this isn't following the runbook anyway, but I was thinking we should increment the development release when we change this so we create new development artifacts that are in line with the next release

jaxesn:
We havent actually released 1.21 yet so this will still be our first release.  The eks team changed the patch version they are going to ship so we are matching.

kschumy (approved):
/lgtm

PR #429, opened 2021-07-07
Changed table generation to use prev release

kschumy:
*Issue #, if available:*

*Description of changes:*
* Added function (`GetComponentVersionsTableWithSameComponentVersionsAsPrevRelease`) to generate components version table for new release by using the previous patch release. 
  * This change is needed because the only existing means of generation (`GetComponentVersionsTable()`) requires the release manifest to exist, which would make the workflow for the bot cumbersome. 
  * This PR does not remove the function to use the manifest, as this is still helpful in manual releases where there could be changes to the components versions.
* Made the generation of the table use the previous patch release's table as the default means of generation. Again, this change was made with the bot in mind and what makes sense for its workflow.
* Minor refactoring and improved comments. For some reason, GitHub makes the changes to existing features, especially in `component_table.go`, look waaaay more intense than they are.

### DIFF
--- a/cmd/release/docs/internal/component_table.go
+++ b/cmd/release/docs/internal/component_table.go
@@ -3,10 +3,8 @@ package internal
 import (
 	. "../../internal"
 	"bytes"
-	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"regexp"
 	"sort"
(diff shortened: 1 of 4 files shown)


cmd/release/docs/internal/component_table.go
```diff
@@ -38,39 +44,107 @@ func GetComponentVersionsTable(release *Release) (string, error) {
...
-	//	where NAME and VERSION are the values needed.
+	// Assumes format for desired lines is "uri: <ecrBase>[...]", and everything from ecrBase to end should be captured.
+	captureRegexForLine := regexp.MustCompile(fmt.Sprintf(`uri: (%s.*)`, ecrBase))
+	foundMatches := captureRegexForLine.FindAllSubmatch(body, -1)
+
+	// Assumes uri format is "<ecrBase>/[some_stuff]/<NAME>:v<VERSION>-<uriSuffix>", where NAME and VERSION are targeted.
+	// Example: "public.ecr.aws/eks-distro/etcd-io/etcd:v3.4.15-eks-1-20-2" -> NAME is "etcd" and VERSION is "3.4.15"
 	captureRegexForName := regexp.MustCompile(`.*/(.*):v`)
-	captureRegexForVersion := regexp.MustCompile(`:v(.*)-eks`)
+	captureRegexForVersion := regexp.MustCompile(fmt.Sprintf(`:v(.*)-%s`, uriSuffix))
```
vivek-koppuru, line 56: Are we anticipating it not be eks anymore?

cmd/release/docs/internal/component_table.go
```diff
@@ -38,39 +44,107 @@ func GetComponentVersionsTable(release *Release) (string, error) {
...
+
+	return strings.Join(tableRows, linebreakString), nil
+}
 
-		tableRows = append(tableRows, tableRow)
+// GetComponentVersionsTableWithSameComponentVersionsAsPrevRelease returns Markdown table of presumed component versions
+// for the release.
+// The previous patch release's component version table must exist in the expected file and in the expected table format.
+// Returned table uses the same components and their versions as the previous patch release. The returned table is not
+// checked for accuracy or compared with the release manifest. References to the EKS-D release number are updated.
```
vivek-koppuru, line 87: I get that this is being done for the purpose of the bot, but why can it not be compared with the release manifest for accuracy? What is our confidence going with this route then?

kschumy:
Vivek and I hoped on a call and talked about a different way to approach this. The revision I just pushed reflects these changes. Now the generation uses the previous release manifest instead of parsing the previous markdown file

vivek-koppuru (approved):
/lgtm

PR #450, opened 2021-07-15
when validating dns wait for 8 successfully resolves before moving on

jaxesn:
From time to time we have a dns issue in prow when creating the kops cluster.  We have a number of retrys in place throughout the script to try and work around this and this validate_dns script was added to give the cluster more time being up to make sure that the dns is in good shape.  Tweaking the script to wait until there are 8 successfully resolves before moving on.

/hold


TerryHowe (approved):
/lgtm

PR #457, opened 2021-07-22
kops: only delete clusters older than a day

jaxesn:
Now that we are going to run more than 1 at a time, we cant be blindly cleaning up.


TerryHowe (approved):
/lgtm

PR #481, opened 2021-08-03
Removed environment and fix num bug

kschumy:
*Issue #, if available:*

*Description of changes:*
* Concentrated `environment` under scope of Release.
	 * `environment` could be both in some cases and nothing outside of Release really needed to know that.
	 * Got rid of `--environment`flag for doc, as there is no need for the doc to know about this and giving that option only create potential problems by requiring the user to know how the generation works.
* Fixed bug if overrideNumber was 0. 
	* There will never be a use case for it to be 0.
	* Allowing it to be 0 let `previousNumber` be -1, which was causing a lot of trouble to deal with.
* Returned Release instead of a pointer because I [read](https://medium.com/@philpearl/bad-go-pointer-returns-340f2da8289) the memory management is better like this in Go.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,4 +153,3 @@ release-docs:
 	go vet ./cmd/release/docs
 	go run ./cmd/release/docs/main.go \
 		--branch=$(RELEASE_BRANCH) \
-		--environment=$(RELEASE_ENVIRONMENT)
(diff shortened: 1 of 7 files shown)


TerryHowe (approved):
/lgtm

PR #495, opened 2021-08-06
coredns/etcd: build oci targets

jaxesn:
CAPI's image-builder will try to embed coredns and etcd into the final image it produces if there is an oci tar in the kubernetes directory next to the kube component tars.


abhay-krishna (approved):
/lgtm

PR #507, opened 2021-08-09
kops: add simple csi and etcd tests

jaxesn:
This is by no means a complete suite, but while switching us over to building minimal base images, it would be great to at least have something that launches our containers and validates they do something right.

The deployment yamls are pulled from upstream.


TerryHowe (approved):
/lgtm

PR #510, opened 2021-08-09
Explicitly disable CGO when building

abhay-krishna:
Disabling CGO explicitly when building etcd binary.


jaxesn (approved):
/lgtm

PR #516, opened 2021-08-13
kops: generate key before running kops

jaxesn:
For 1.21 we are running amd64 and arm tests concurrently.  Kops will generate an ssh key if one doesnt exist, but this creates an issue for which ever cluster starts the process second since it tries to overwrite the key that was just created.  This creates the key before calling kops.  Also, store the e2e results in separate folders.


TerryHowe (approved):
/lgtm

PR #518, opened 2021-08-17
support multiple 3.x pause tags and set 3.4.1 for 1.20

jaxesn:
kubeadm 1.21 hardcodes 3.4.1 as the pause tag, we need to push this in addition to 3.4 for 1.20.


abhay-krishna (approved):
/lgtm

projects/kubernetes/kubernetes/Makefile
```diff
@@ -5,7 +5,7 @@ RELEASE?=$(shell cat $(BASE_DIRECTORY)/release/$(RELEASE_BRANCH)/$(RELEASE_ENVIR
 ARTIFACT_BUCKET?=my-s3-bucket
 GIT_TAG?=$(shell cat ./$(RELEASE_BRANCH)/GIT_TAG)
 GOLANG_VERSION?=$(shell cat ./$(RELEASE_BRANCH)/GOLANG_VERSION)
-PAUSE_TAG?=$(shell cat ./$(RELEASE_BRANCH)/PAUSE_TAG)
+PAUSE_TAG?=$(shell cat $(wildcard ./$(RELEASE_BRANCH)/PAUSE_*TAG))
```
TerryHowe, line 8: I don't have this command on my mac

abhay-krishna: This is a Make function, so it will be recognized only within the scope of the Makefile.

TerryHowe: Thanks. Make is so weird

PR #519, opened 2021-08-17
Fix regex in main postsubmits

abhay-krishna:
Building all projects must be scoped only to the Makefile at the root.


jaxesn (approved):
/lgtm

PR #525, opened 2021-08-19
kops: use api server ip instead of relying on dns

jaxesn:
In the postsubmits we often run into a seemingly random issue resolving the ip from domain name for the new cluster.  It seems like a dns issue, this will force the api calls to go over IP instead.


abhay-krishna (approved):
/lgtm

PR #535, opened 2021-08-19
install kops/kubectl before running prow.sh

jaxesn:
In 1-21 postsubmit we run both amd and arm clusters.  They both try to download kops and some times, happened last night, it looks like one overwrites the file while the other is trying to use it resulting in: `./create_configuration.sh: line 51: bin/kops: Text file busy`


TerryHowe (approved):
/lgtm

TerryHowe (approved):
/lgtm

PR #541, opened 2021-08-22
kops: cleanup only old clusters with same arch

jaxesn:
Since 1-21 run both arm and amd tests, if there is a situation where it needs to cleanup old clusters, they would both try to clean up the same clusters causing issues.  This tweaks the cluster name and check when cleaning up to only touch clusters that are the same version and arch.


TerryHowe (approved):
/lgtm

PR #693, opened 2022-01-22
coredns + etcd: fix upload location for oci tars

jaxesn:
*Issue #, if available:*

*Description of changes:*
etcd and coredns oci tars are supposed to be put in the kubernetes/<version>/bin/linux/amd64 directory to match the expectation upstream and by image-builder used in eks-a.  This was broken and instead these files were being uploaded to the etcd/bin + coredns/bin folder instead.  The code was already setup to handle tars and oci tars differently and there were already separate validate_images + validate_artifacts files for these two projects.  The miss was in the second call to copy_artifacts in Common.mk was passing the REPO, coredns or etcd, instead of kubernetes to force the directory location.


abhay-krishna (approved):
/lgtm

PR #701, opened 2022-02-02
set 1.22 builds to use the minimal base image

jaxesn:
*Issue #, if available:*

*Description of changes:*

Starting with 1.22 we want to use the minimal base images for all containers.  This removes the temporary code we had to support building 1.21 with both and testing.


kschumy (approved):
/lgtm

PR #705, opened 2022-02-03
etcd: update to 3.5.1 for 1.22

jaxesn:
*Issue #, if available:*
For 1.22 we are bumping to use etcd 3.5.x from 3.4.

*Description of changes:*
There is a lot going on here.  The largest changes are in Common.mk and add support for building binaries in the same project from different folders, each with their own go.mod.  In etcd 3.5 they separated the go.mod files between "server" and "etcdctl.  This requires gathering licenses and building the attribution separately for the two.  We had a way to handle this before but it was quite manual, see https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/kubernetes-sigs/cluster-api/Makefile#L24 for an example.  The changes here abstract as much of it away as possibly and keep the project makefile cleaner.

This also introduces a new concept we have ran into before where etcd 3.5 is built slight differently than 3.4.  I have introduced an Overrides.mk file to the release branch folder for 1-22 which is included in the main etcd makefile.  In this new file we can override the things to support the differences between building the two verisions.

/hold


Common.mk
```diff
@@ -307,6 +323,106 @@ define CGO_BINARY_TARGET_BODY
...
+define GATHER_LICENSES_ALL_TARGETS_BODY	
+	$(eval $(foreach binary,$(1), $(call GATHER_LICENSES_TARGETS_BODY,$(binary),$(2))))
+endef
+
+# $1 - binary file name
+# $2 - go mod path for binary
+define GATHER_LICENSES_TARGETS_BODY
+	$(call LICENSE_TARGET_FROM_BINARY_GO_MOD,$(1),$(2)): $(REPO)/$(2)/eks-distro-go-mod-download
+		$$(BASE_DIRECTORY)/build/lib/gather_licenses.sh $$(REPO) $(MAKE_ROOT)/$(call LICENSE_OUTPUT_FROM_BINARY_GO_MOD,$(1),$(2)) "$$(LICENSE_PACKAGE_FILTER)" $(2)
+
```
abhay-krishna, line 354: do we need this extra line here?

jaxesn: Yeah, not sure why but any of these "functions" which are evald like this need the trailing newline

abhay-krishna (approved):
/lgtm

PR #709, opened 2022-02-04
Pass cluster name as flag instead of positional argument

abhay-krishna:
We set `KOPS_CLUSTER_NAME` as an env var, which gets set as the [default value for the `--name` flag](https://github.com/kubernetes/kops/blob/v1.22.3/cmd/kops/root.go#L138-L139). So passing it as a positional argument makes kOps see the value as coming from two places, so it fails.

Note: Just removing the cluster name (positional arg) from the command would also work, but explicitly specifying it with `--name` helps us track exactly what values were passed.


jaxesn (approved):
/lgtm

PR #710, opened 2022-02-04
fix 1-22 dev release build and kops script

jaxesn:
*Issue #, if available:*

*Description of changes:*
- The 1-22 builds were getting overwritten by the dev-release build settings and using the non-minimal images.  This should fix that
- Tweaked the kops script to better handle the new version esp wrt the older kube versions.

The kops roles changes to add the newly required perm has been rolled out to prod so we should see passing conformance builds for at least 1.18-1.21.

Testing

Created 1.18-1.21 clusters with kops locally and they all came up correctly.

### DIFF
--- a/development/kops/cluster_wait.sh
+++ b/development/kops/cluster_wait.sh
@@ -34,24 +34,21 @@ do
     echo 'Waiting for cluster to come up...'
 done
 
-# In kops 1-21 the metrics addon is not configurable enough for 1.18 based clusters
-# manually add -kubelet-insecure-tls
-# https://github.com/kubernetes/kops/blob/v1.21.0-beta.3/upup/models/cloudup/resources/addons/metrics-server.addons.k8s.io/k8s-1.11.yaml.template#L140
-# UseKopsControllerForNodeBootstrap is only true when 1.19 and above
-if [ "${RELEASE_BRANCH}" == "1-18" ]; then
-    PATCH='[{"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--kubelet-insecure-tls" }]'
-    while ! kubectl --context $KOPS_CLUSTER_NAME  -n kube-system patch deployments metrics-server --type=json -p="$PATCH"
-    do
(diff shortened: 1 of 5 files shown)


development/kops/cluster_wait.sh
```diff
@@ -34,24 +34,21 @@ do
     echo 'Waiting for cluster to come up...'
 done
 
-# In kops 1-21 the metrics addon is not configurable enough for 1.18 based clusters
```
jaxesn, line 37: This appears to be not the case in kops 1-22 so I removed this block.

release/prow.sh
```diff
@@ -22,10 +22,12 @@ cd ${BASE_DIRECTORY}
...
 
 export RELEASE=$(cat ${BASE_DIRECTORY}/release/${RELEASE_BRANCH}/${RELEASE_ENVIRONMENT}/RELEASE)
-export KUBE_BASE_TAG=$(cat ${BASE_DIRECTORY}/projects/kubernetes/release/${RELEASE_BRANCH}/GIT_TAG)-eks-${RELEASE_BRANCH}-${RELEASE}
-export PRESUBMIT_KUBE_BASE_TAG=$(cat ${BASE_DIRECTORY}/projects/kubernetes/release/${RELEASE_BRANCH}/GIT_TAG)-eks-${RELEASE_BRANCH}-${RELEASE}
-export BASE_REPO=${BASE_REPO:-${IMAGE_REPO}}
-export BASE_IMAGE=${BASE_IMAGE:-public.ecr.aws/eks-distro-build-tooling/eks-distro-base:$(cat ${BASE_DIRECTORY}/EKS_DISTRO_BASE_TAG_FILE)}
+
+if [ "${RELEASE}" == "0" ]
+then
+    echo "No production release zero"
```
abhay-krishna, line 28: nit:
```suggestion
    echo "No dev release zero"
```
but IDC

jaxesn: lol

PR #715, opened 2022-02-10
for postsubmit jobs use dev release + 1 as pre release

jaxesn:
*Issue #, if available:*

*Description of changes:*
Postsubmits and dev release jobs write to the same location on s3 and use the same container image tags in ecr.  This hasnt been an issue but we are now trying to use the dev release for testing and like to have a stable release which does not change.  

This changes introduces the concept of a `pre` release.  For the postsubmit job, the RELEASE var is changed to be what would be the next dev release, but with a trailing .pre.  Changing the RELEASE var should change both the location on s3 and the image tags pushed to ecr.  

Notes:

- A non-numeric release would not actually be allowed to run through the release tooling since the CRD specifics release as a number. This shouldn't be an issue because post submits do not run the release tooling.
- Adding `.pre` to the release could put us over the limit in terms of url length since kops uses the full url to create files on the linux host which have a max length.  We will know pretty quickly if this is the case based on the results of the postsubmit jobs.

/hold

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,7 @@ kops-prereqs:
 	cd development/kops && ./install_requirements.sh
 
 .PHONY: postsubmit-conformance
+postsubmit-conformance: RELEASE:=$(shell echo  $$(($(RELEASE) + 1))).pre
 postsubmit-conformance: postsubmit-build kops-prereqs kops-prow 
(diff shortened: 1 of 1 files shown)


vivek-koppuru (approved):
/lgtm

PR #717, opened 2022-02-11
kubernetes: fix version file validation by remove .pre for postsubmits

jaxesn:
*Issue #, if available:*

*Description of changes:*
In postsubmits we now make the tag `(release+1).pre` to avoid overwriting the artifacts pushed for the actual release build.  This removes the code we used to use to remove the .minimal and replaces it with removing the .pre for the kube-verison-file validation.


vivek-koppuru (approved):
/lgtm

PR #719, opened 2022-02-11
Deleted kms-provider-0 logging patch

kschumy:
*Description of changes:*
* Delete kms-provider-0 logging patch for all branches. 
  * We decided that this wasn't great for users, as they would not know about errors related to kms-provider-0 without setting up a separate alarm

____


jaxesn (approved):
/lgtm

PR #721, opened 2022-02-13
Remove remaining kube git version file references and fix broken release generation

kschumy:
*Description of changes:*
* Follow up to this PR #718, where we remove `KUBE_GIT_VERSION_FILE`. This PR cleans up some of the remaining code related to this file.
* The release number generation broke after the removal of `KUBE_GIT_VERSION_FILE`, so this PR also fixes this issue.

---


projects/kubernetes/kubernetes/Makefile
```diff
@@ -96,10 +95,6 @@ test: $(BINARY_TARGETS)
 fix-licenses:
 	build/fix_licenses.sh $(REPO) $(OUTPUT_DIR)
 
-.PHONY: create-version-file
```
jaxesn, line 99: We still need to make the file because the release tooling uses it.

jaxesn (changes requested):

projects/kubernetes/kubernetes/Makefile
kschumy: So we need to make it but not have it in the repo?

jaxesn: correct

kschumy:
Closing this PR to merge it's some of the changes into PR #722, which has to do with fixed the now broken make commands for cutting releases.

I'm going to keep the commit "[Removed KUBE_GIT_VERSION_FILE code for releases](https://github.com/aws/eks-distro/pull/721/commits/581410658f331dbf3b30e55fe435f995ad57ba6b)" because this has to do with fixing the broken commands.

I'm going to drop the commit "[Removed KUBE_GIT_VERSION_FILE related code](https://github.com/aws/eks-distro/pull/721/commits/ce885e5e865564a0bfe2f4aea7bea659f543a980) " because this isn't needed, according to Jackson's comment on this PR.

PR #722, opened 2022-02-14
Fix and major refactor of make for releases

kschumy:
*Issue #, if available:*

*Description of changes:*
This PR fixes two bugs that broke the make commands for docs and changing release numbers. It also changes some default flag values to ones that make more sense or typical workflow

I've run the commands to test them and confirm they work.


jaxesn (commented):
This looks pretty reasonable to me.  Maybe some minor cleanup left to do.

I wonder if we could add a presubmit job to run this in a dry run capacity to catch if changes would break this flow?

jaxesn (approved):
/lgtm

PR #738, opened 2022-02-17
kops: handle kubectl download failure and amd/arm overwriting

jaxesn:
*Issue #, if available:*

*Description of changes:*
There we two postsubmit failures earlier today.

- We have seen this a number of times before and I believe its due to multiple postsubmits running at once, but sometimes kubectl fails to download from s3.  I believe this is a timing thing where one job is updating it and another is trying to download at the same time...  
```
/home/prow/go/src/github.com/aws/eks-distro/development/kops/bin/kubectl: line 1: syntax error near unexpected token `newline'
/home/prow/go/src/github.com/aws/eks-distro/development/kops/bin/kubectl: line 1: `<?xml version="1.0" encoding="UTF-8"?>'
```
This PR adds a retry loop making sure it gets downloaded correctly before continuing.

- We have also the other one that has to do with the arm and amd tests, which only run for 1.21+, trying to update the kubeconfig at the same time.  

`Error: open /root/.kube/config.lock: file exists`
This PRs adds a 5m sleep before kicking off the arm tests to try avoid these overlaps.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,7 @@ kops-prow-arm: export NODE_ARCHITECTURE=arm64
 kops-prow-arm: postsubmit-build
 	$(eval MINOR_VERSION=$(subst 1-,,$(RELEASE_BRANCH)))
 	if [[ $(MINOR_VERSION) -ge 21 ]]; then \
+		sleep 5m; \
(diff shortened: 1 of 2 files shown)


kschumy (approved):
/lgtm

PR #740, opened 2022-02-18
kops: run prereqs after build is complete

jaxesn:
*Issue #, if available:*

*Description of changes:*
The reason some builds were failing is because the install-prereq script downloads kubectl based on the current build.  If this is running for the first time, kubectl wont exist yet so it will retry and then eventually fail causing the job to be killed.  This sets the build target as a prereq for install-prereqs which should avoid this case.


abhay-krishna (approved):
/lgtm

PR #753, opened 2022-02-24
kube: use short commit hash after applying patches as binary version

jaxesn:
*Issue #, if available:*

*Description of changes:*

When we originally shipped EKS-D, we tagged kube images as `v1.21.5-eks-1-21-1` and we also used this as the version for the binaries.  This meant that if you ran `kubectl get nodes -owide` you'd see that as the node version.  Container images continue to be tagged this way.

When adding the reproducible builds and checksums, we dropped the release number from the version that is embedded in the binary, so its just `v1.21.5-eks-1-21` (which is admittedly not as good) to avoid having to constantly update the checksums of the binaries with every release since most releases are just image bumps with a new release number.  As a reference, when running get nodes against an EKS cluster and their kubelet returns `v1.21.5-eks-<some short git hash>`.   

We have recently changed the way the patches are being applied during the build to use `git am --committer-date-is-author-date` which makes it so that the git commit/timestamps that are generated is actually consistent between builds, assuming patch content hasnt changed.

This PR changes the binary versions to be `v1.21.5-eks-<short hash>` and that short hash will always be the same if the patch content hasn't changed.  If the patches change, that would trigger a new hash and a new checksum anyway, since its a new build.  With this approach, we could go back to be able to actually map a given kubelet to a specific release of EKS-D, which since dropping the release number we cant really do, and the kubelet version would also be something to signify to the user that there were actually build/patch changes in a given EKS-D release.

This also changes the tag we create before generating the source to include this short hash.  This means that downstream users of this tar (BR), when they build from the source package, the version embedded into the binary will be the same as EKS-D.  This means on the EKS-A, there would be no version difference between running get nodes against a Ubuntu cluster and a BR cluster.

We will also need to make this change to the release tooling to generate the image tag instead of relying on the version file: https://github.com/aws/eks-distro-build-tooling/pull/316

/hold


chrisdoherty4:
Food for though: commit hashes often better compliment the `+` portion of a semver string reserving the `-` portion to communicate release candidate, beta, alpha etc information. E.g. `v1.21.1-rc1+<short sha>` Is it worth discussing a switch?

jaxesn:
Yeah this comes up from time to time on the EKS side.  Customers have brought it up before in relation to it causing issues for helm charts and in other areas, but they have decided to stick with the - approach.  I want to continue to match them.  Ill check in with them and make sure they havent had more recent discussions around this topic.

kschumy (approved):
/lgtm

PR #756, opened 2022-03-02
Updated projects for 1-22

kschumy:
*Description of changes:*
* Changed **ONLY** the project versions for 1.22. Everything else remains the same version as it was before this change. A number of project directory were split into release branches because of this.

#### N.B., There are a number of projects that should be updated to newer patch/minor versions for some release branches. Those will be handled in later PRs.
 
----


kschumy:
/hold

Jobs likely need updating

jaxesn (approved):
Ive doubled checked all the versions here, the only changes are on the 1-22 versions.  Most of these changes are all the same change breaking the 5 remaining csi projects which werent released branch'd into release branches.  This required makfile/dockerfile/build scripts/etc changes in addition to all the standard files that need to exist for each release branch.  The PR seems bigger than it really is because of this.

/lgtm

PR #766, opened 2022-03-08
kube: fix building twice due to path mismatch

jaxesn:
*Issue #, if available:*

*Description of changes:*
Kubernetes was building twice because the default output path is `linux-amd64`, but kube was building to `linux/amd64` and I *think* always has been.  This overrides the binary_target to avoid this.  The dockerfile and everything else was already pointing to the correct `linux/amd64` paths.


TerryHowe (approved):
/lgtm

PR #767, opened 2022-03-08
Changed Kubernetes version to 1.20.15 and updated patches

kschumy:
*Issue #, if available:*

*Description of changes:*
* Removed unneeded patches
* Removed some lines in patches that we're apparently no longer going to update


kschumy:
Checksum will fail. I'll update them

projects/kubernetes/kubernetes/1-20/patches/0002-EKS-PATCH-Pass-region-to-sts-client-for-ap-east-1.patch
```diff
@@ -119,6 +119,3 @@ index fbbaa6567d0..6f280c19320 100644
  	regionName, err := azToRegion(zone)
  	if err != nil {
  		return nil, err
```
TerryHowe, line 121: Are making these with `git format-patch` or did you manually just remove this trailer (which I thought was the git version)? Same with the  "From" I thought was the previous commit, does that make it a wild card?

kschumy: We're doing something different for this version. I'll slack you

jaxesn (changes requested):

jaxesn (approved):
/lgtm

PR #775, opened 2022-03-10
kops test: switch back to kops managed cilium and fix session affin

jaxesn:
*Issue #, if available:*

*Description of changes:*

This switches back to the kops managed cilium and settings enable session affinity to fix the 3 failing conformance.  From local testing, I believe this configuration should show passes for all the ipv4 runs and then the ipv6/arm 1.21 cluster *should* pass and then the 1.22 one should fail due to the bad kube-proxy. 🤞  


abhay-krishna (approved):
/lgtm

PR #791, opened 2022-03-15
Updated CSI snapshotter for 1-20 and 1-21

kschumy:
*Description of changes:*
* Bumped CSI Snapshotter for 1-20 and 1-21 from v4.1.1 to v4.2.1
  * This decision was made in an effort to update dependencies of CSI Snapshotter. The updated version is the newest minor version  available that supports these two Kubernetes versions. See the [supported version table](https://kubernetes-csi.github.io/docs/external-snapshotter.html#supported-versions).
  * Golang is [not updated](https://github.com/kubernetes-csi/external-snapshotter/blob/v4.2.1/release-tools/prow.sh#L89) in v4.2.1
 
---


jaxesn (approved):
/lgtm

PR #795, opened 2022-03-15
fix versions for metrics server and iam auth and add validation for v…

jaxesn:
…ersion

*Issue #, if available:*

*Description of changes:*
There was a bug due to the iam-auth binary not having the proper ldflag set for the version.  The package is super important here and must match the upstream. This also adds a validation to the build phase to make sure its what we expect it to be in case these change in the future.  I also updated the readmes some to call out more specific places to check for this kind of thing.


kschumy (approved):

kschumy (approved):
/lgtm

PR #848, opened 2022-04-02
update go versions in install helper and fix typo in common.mlk

jaxesn:
*Issue #, if available:*

*Description of changes:*

This brings the versions inline with whats in builder base in case people use this locally for setting up their machines.  Switches to go install to remove deprecation warning about using go get. Cleans up the goversion path handling since we never pass in a go version with a patch version none of this should have been necessary. 

### DIFF
--- a/Common.mk
+++ b/Common.mk
@@ -472,7 +472,7 @@ $(KUSTOMIZE_TARGET):
 	curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash -s -- $(OUTPUT_DIR)
 
 .PHONY: clone-repo
-clone-rpeo: $(REPO)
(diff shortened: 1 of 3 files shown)


TerryHowe (approved):
/lgtm

Common.mk
```diff
@@ -472,7 +472,7 @@ $(KUSTOMIZE_TARGET):
 	curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash -s -- $(OUTPUT_DIR)
 
 .PHONY: clone-repo
-clone-rpeo: $(REPO)
+clone-repo: $(REPO)
```
abhay-krishna, line 475: Are we gonna let the bot do the help.mk update?

PR #861, opened 2022-04-13
kubernetes: avoid out of space in presubmit by using mv

jaxesn:
*Issue #, if available:*

*Description of changes:*

Space in presubmit is super tight. I think with our recent changes to the garbage collection settings of buildkit, we are hitting the limit right at the end of the build process when copying the binaries to the final artifacts directory.  This should fix that.


kschumy (approved):
/lgtm

PR #931, opened 2022-05-12
snapshotter: Fixes go.sum false positive alert for protobuf def

jaxesn:
The previous PR which fixes this did not actually remove the bad version
from the go.sum file.

Funny story, go.sum is not really a list of
deps included in the actually binary, its more of a manifest that go
used to walk the dep tree. Dependabot will sometimes improperly use this
as the source of truth which is not 100% correct. I am not 100% sure
if that is what is happening here, but it def could be because the final
binary is only including 1.3.2 of protobuf.

The reason this wasnt cleaned up is that go 1.15's `go mod tidy` does
not do as good of job of cleaning up the go.sum as 1.16.  I have modified
my patching scripts to delete the go.sum along with the vendor dir and regenerate
it instead.

ref: https://github.com/dependabot/dependabot-core/issues/4740

*Issue #, if available:*

*Description of changes:*


drewvanstone (approved):
LGTM

jeffwidman:
BTW, GitHub fixed it so `go.sum` is no longer used as a source of dependencies: https://github.blog/changelog/2023-03-07-dependency-graph-removes-go-sum-support/

PR #933, opened 2022-05-12
Initial changes for 1.23

kschumy:
*Description of changes:*
* Added projects (except etcd, which Jackson already added)
  * Did NOT add Kubernetes patches because we're still figuring those out
* Update version conditional statements throughout


kschumy:
checksums will fail. I'll fix them

projects/kubernetes/release/1-23/GIT_TAG
```diff
@@ -0,0 +1 @@
+v0.12.0
```
kschumy, line 1: There is a newer version of this ((v0.13.0)[https://github.com/kubernetes/release/tree/v0.13.0]), but I thought we might want to just stay on where we were for 1.22. It wasn't super clear from upstream (https://github.com/kubernetes/kubernetes/blob/v1.23.6/build/common.sh#L94)

jaxesn: I'll look at this one with you

jaxesn (commented):
Do we need to add 1-23 to that supported releases file? Or do we not do that until we ready for prod release?

kschumy:
> Do we need to add 1-23 to that supported releases file? Or do we not do that until we ready for prod release?
In the past, we did that as a separate PR, but I added it

jaxesn (approved):
/unhold
/approve

PR #935, opened 2022-05-13
Fixes go-runner path in docker container

jaxesn:
For 1.23 clusters kops changed the static pod manifests to run
/go-runner instead of relying on the entry point set in the dockerfile.
go-runner has been at the / of the upstream container for many versions now.
This should continue to work with previous versions as well

Kops PR:  https://github.com/kubernetes/kops/pull/12494

*Issue #, if available:*

*Description of changes:*


projects/kubernetes/release/1-23/GIT_TAG
```diff
@@ -1 +1 @@
-v0.12.0
+v0.13.0
```
jaxesn, line 1: this doesnt materially matter, but this version is probably more correct.

markapruett (approved):
/lgtm

PR #937, opened 2022-05-13
Fixes kube-proxy for 1.23

jaxesn:
Same go-runner fix as in go-runner base image.

I am not sure what is up with the metrics server thing for 1.23.
From the looks of the kops source it should be the same as 1.22
but for some reason metrics server can not bind to port 443.  This
changes treats 1.23 like 1.19 and 1.20 to see what happens.

I think we can remove this if altogether, but we can test that out
after fixing 1.23

*Issue #, if available:*

*Description of changes:*


projects/kubernetes/kubernetes/1-23/ATTRIBUTION.txt
```diff
@@ -2426,3 +2426,4 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ------
 
+
```
mitalipaygude, line 2429: nit: unnecessary new line?

jaxesn: i want to force the kube project to build

PR #1048, opened 2022-06-21
Cleans up go.mod/sum files

jaxesn:
*Issue #, if available:*

*Description of changes:*

We are still going thru these to validate the files being added are correct and the alerts are legit and not full of false positive both due to checking in the wrong versions or alerts that do not apply.  We are going to start going thru these 1 version at a time to reduce this noise and validate along the way.


kschumy (approved):
/lgtm

PR #1165, opened 2022-08-03
build and install kops from source with patch removing req on flannel…

danbudris:
… plugin

*Issue #, if available:*

*Description of changes:*
Apply patch to kOps to remove flannel plugin from hardcoded list of CNI plugins.

Do this by cloing kops from the desired tag, patching it and installing it, rather than just curling the release version.

ran `./install_requierments.sh` to test

### DIFF
--- a/development/kops/install_requirements.sh
+++ b/development/kops/install_requirements.sh
@@ -33,13 +33,19 @@ mkdir -p ${BASEDIR}/bin
 if [ ! -x ${KOPS} ]
 then
     echo "Determine kops version"
-    KOPS_VERSION="v1.23.2"
-    echo "Download kops"
-    KOPS_URL="https://github.com/kubernetes/kops/releases/download/${KOPS_VERSION}/kops-${OS}-${ARCH}"
-    set -x
-    curl -L -o ${KOPS} "${KOPS_URL}"
-    chmod 755 ${KOPS}
-    set +x
+    KOPS_VERSION_TAG="v1.23.2"
+
+    echo "Cloning kops"
+    KOPS_FLANNEL_PLUGIN_PATCH="0001-remove-hardcoded-requierment-on-flannel-plugin.patch"
+    KOPS_GIT_URL="https://github.com/kubernetes/kops.git"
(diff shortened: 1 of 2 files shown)


development/kops/install_requirements.sh
```diff
@@ -33,13 +33,19 @@ mkdir -p ${BASEDIR}/bin
...
+    KOPS_FLANNEL_PLUGIN_PATCH="0001-remove-hardcoded-requierment-on-flannel-plugin.patch"
+    KOPS_GIT_URL="https://github.com/kubernetes/kops.git"
+    git clone ${KOPS_GIT_URL}
+
+    echo "Patching kops"
+    git -C ${BASEDIR}/kops checkout ${KOPS_VERSION_TAG} 
+    git -C ${BASEDIR}/kops am ../patches/${KOPS_FLANNEL_PLUGIN_PATCH}
+
+    echo "Building kops"
+    (cd ${BASEDIR}/kops && make kops-install)
```
jaxesn, line 48: where does the kops binary go?

danbudris: $GOPATH/bin

danbudris: we could set `GOBIN` inline to put it elsewhere

jaxesn: yeah can u just put where we used to download it?

development/kops/patches/0001-remove-hardcoded-requierment-on-flannel-plugin.patch
```diff
@@ -0,0 +1,24 @@
+From 2ec5afbedada5688c13aa8e8d9ac9ffba3ae4511 Mon Sep 17 00:00:00 2001
+From: Daniel Budris <budris@amazon.com>
+Date: Wed, 3 Aug 2022 11:50:10 -0400
+Subject: [PATCH] remove hardcoded requierment on flannel plugin
+
```
kschumy, line 5: Can we add a description here? It would be helpful if we ever have the option of removing this patch with a kops update

jaxesn (approved):

PR #1178, opened 2022-08-24
Initial changes for 1.24

kschumy:
*Issue #, if available:*

*Description of changes:*
* **Project files**
  * For all projects EXCEPT Kubernetes
    * Copied everything we used in 1.23. I did not change anything except CHECKSUM paths. We can iterate on these versions in future releases. We're just trying to get something out the door for now.
  * For Kubernetes
    * Added the patches
    * Set Kubernetes version to [v1.24.4](https://github.com/kubernetes/kubernetes/tree/v1.24.4)
    * Set Golang version to 1.18, which is [what upstream uses](https://github.com/kubernetes/kubernetes/blob/v1.24.4/build/common.sh#L94)
    * Set pause to 3.7 instead of 3.6 like 1.23 used because this appears to be [what upstream did](https://github.com/kubernetes/kubernetes/blob/v1.24.4/build/dependencies.yaml#L149)
    * Kept release the same as 1.23 because I \*think\* (though I am not 100% sure of this) that's what upstream is doing. It's v0.13.0, but there is a new version v0.14.0. See what Kubernetes uses [here](https://github.com/kubernetes/kubernetes/blob/v1.24.4/build/common.sh#L93-L95) and [here](https://github.com/kubernetes/kubernetes/blob/v1.24.4/build/dependencies.yaml) vs what's in the [releases versions](https://github.com/kubernetes/release/releases)
* **Other files**
  * Added `1-24` to `MINIMAL_VARIANT_VERSIONS` in Common.mk (we probably could get rid of this since they all use minimal base image now)
  * Added `1-24` to `release/SUPPORTED_RELEASE_BRANCHES`
  * Created `release/1-24/production/RELEASE` and `release/1-24/development/RELEASE` and set both to `0`


jaxesn:
It is hard to tell which version of the release project goes with what kube verision. I think we are safe to bump to the 0.14 tag which will then allow us to bump to golang 1.18 for the go-runner build, which would be nice and i suspect what they did in 1.24 since kube itself is built with 1.18.

We can go ahead and merge this and do that in a follow up

jaxesn (approved):

kschumy:
bumping k8s release to 1.14 and its go to 1.18 broke a lot. I undid it and am seeing what will happen

projects/kubernetes/kubernetes/build/run_tests.sh
```diff
@@ -44,10 +54,10 @@ export PATH="${GOPATH}/bin:${MAKE_ROOT}/kubernetes/third_party/etcd:${PATH}"
 MAX_RETRIES=5
 # There are flakes in upstream tests, the process also caches passing results
 # so on rerun it only runs the tests which failed
-for i in $(seq 1 $MAX_RETRIES); 
-do 
+for i in $(seq 1 $MAX_RETRIES);
+do
   [ $i -gt 1 ] && sleep 5
-  make test && ret=0 && break || ret=$?; 
+  make test && ret=0 && break || ret=$?;
```
kschumy, line 60: IntelliJ reformatted this on its own. This change does not matter for the 1.24 release

kschumy:
There are issues with updating kops. We're waiting on those to be fixed before moving forward with this PR

jaxesn (approved):
/approve
/unhold

PR #1257, opened 2022-09-27
Adds buildkit cache to image builds

jaxesn:
*Issue #, if available:*

*Description of changes:*

Weve been doing this for a while on the eks-a side, but never retrofitted eks-d to do the same.

Buildkit allows embedding the "build cache", I believe this is basically just the hash of the content for a given layer, into the final image which can then be imported on subsequent builds. If on the subsequent build the hash of the content for a given layer matches what is in the previous image's cache, buildkit will reuse the same layer vs building it again, which would result in a new layer id/sha.  This is basically what would happen if you build the same image over and over on the same buildkit builder, each subsequent build would finish almost instantly since its able to reuse the cache.

We should benefit from this since when we cut new releases, rarely are all component changed, which means between releases if etcd, for ex, did not change, the resulting images across these releases would end up being the same exact image, with multiple tags.


danbudris (approved):
/lgtm

PR #1242, opened 2022-09-06
Major refactor of release make commands

kschumy:
*Issue #, if available:*
closes #743 
closed #972 

*Description of changes:*
Major refactor of make docs and make release number. From a user standpoint, there are few noticeable differences. But a ton of fragile and confusing confusing code was removed. Most of it dealt with handling hypothetical use cases that simply are not needed.

I tested the changes for all happy paths, and they worked. The next step will be to add more undo ability, but this change is not critical at this time. We need this changed merged before this is done so we can unblock Cameron, who is working on adding the K8s version as part of the docs generation process.

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1356, opened 2022-10-24
Fixed make github-release and added tests

### LINKED ISSUE #1355: Fix override number for make github-release
**What would you like to be added**:
A previous change removed the override number option from Release because it was thought this was not needed and the code implementing this a mess. But the `make github-release` command still needed this value and it broke without it. (FOLLOW UP: make docs isn't broken but it does need this feature added to it. See ticket https://github.com/aws/eks-distro/issues/1354)

**Why is this needed**:
I broke things.

kschumy:
*Issue #, if available:*
closes #1355

*Description of changes:*
* Changes to Release
  * Moved Release to it's own directory, which is why it looks new
  * Added override number option when creating Release (DO NOT USE THIS IN MOST CASES)
  * Created releaseinput.go to handle internal logic of creating Release
  * Created tests for Release
* Added override number option to github-release

*Confirmation that it worked* 
* `RELEASE_BRANCH=1-20 OVERRIDE_NUMBER=21 make github-release` generated this https://github.com/aws/eks-distro/releases/tag/v1-20-eks-21
* `RELEASE_BRANCH=1-21 make github-release` generated this https://github.com/aws/eks-distro/releases/tag/v1-21-eks-19

---


cmd/release/gh-release/main.go
```diff
@@ -26,18 +26,30 @@ var (
 // TODO: implement number override for release
 func main() {
 	branch := flag.String("branch", "", "Release branch, e.g. 1-22")
+	overrideNumber := flag.String("overrideNumber", "", "Optional override number, e.g. 1")
```
markapruett, line 29: Should this be named something like "override version" to distinguish that it is a string and not a number?

kschumy: Release number is an established thing throughout the code and release process. I would be hesitant to change this to version because I can imagine that there would be confusion between this and Kubernetes version.

Makefile
```diff
@@ -217,7 +218,7 @@ release-docs:
 
 .PHONY: github-release
 github-release:
-	go vet ./cmd/release/github_release
-	go run ./cmd/release/github_release/main.go \
+	go vet ./cmd/release/gh-release
+	go run ./cmd/release/gh-release/main.go \
 		--branch=$(RELEASE_BRANCH) \
-		--number=$(PROD_RELEASE)
```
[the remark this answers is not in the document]
kschumy: It was a thing before the major refactor: https://github.com/aws/eks-distro/pull/1242 

I had removed it because I didn't think we needed it, and the original implementation I did of it was bad. It allowed for way to much flexibility and the logic to deal with all this flexibility was a nightmare to navigate through and maintain

PR #1496, opened 2022-11-29
Change PHONE to PHONY

markapruett:
*Issue #, if available:*

*Description of changes:*
Fix phony target definition.

A couple questions:
1. Was this intentional to turn this off for the time being?
2. I assume this just gets ignored instead of erroring somewhere
3. Has this ever run correctly since it was added?


jaxesn (approved):
Good catch!  I dont think this make any material difference, but we should be correct and this target is clearly a "phony" target.

PR #1533, opened 2022-12-09
Fix layer caching

markapruett:
*Issue #, if available:*
Fixes layer caching

*Description of changes:*
For the import cache during image build check the latest tag in both dev and prod ECRs and use that as the cache.


jaxesn (approved):
/approve

PR #1537, opened 2022-12-12
Builds components first for most post submits

jaxesn:
*Issue #, if available:*

*Description of changes:*

Description in makefile.

### DIFF
--- a/cmd/main_postsubmit.go
+++ b/cmd/main_postsubmit.go
@@ -64,6 +64,7 @@ func main() {
 	gitRoot := flag.String("git-root", "", "Git root directory")
 	dryRun := flag.Bool("dry-run", false, "Echo out commands, but don't run them")
 	rebuildAll := flag.Bool("rebuild-all", false, "Rebuild all projects, regardless of changes present")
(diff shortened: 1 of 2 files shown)


cmd/main_postsubmit.go
```diff
@@ -117,6 +121,13 @@ func main() {
 		"kubernetes-csi/external-snapshotter",
 		"kubernetes-csi/external-provisioner",
 	}
+
```
markapruett, line 124:
```suggestion

# Changing the order here to try to ensure the projects (etcd, metrics-server, csi's etc) are built first. 
# All the release branches build in parallel, and since kubernetes takes a longer time, having the first version 
# build the other projects first, when the remaining release branches get around to building the other 
# projects, they'll have those images ready to pull and use for the layer cache if applicable. Using the 
# cached version, the projects that are the same regardless of release branches (i.e. metric-server)
# should not have to produce a new (duplicate) image and can tag the existing one if it's the same. 
```

jaxesn: I wrote something similar in the makefile. do you want them both? would you rather over what i wrote?

markapruett: I guess it makes sense to just have it in the makefile

markapruett (approved):

PR #1540, opened 2022-12-13
Cleans yum after pause build

jaxesn:
*Issue #, if available:*

*Description of changes:*

To avoid ballooning on presubmit which run on fargate without an overlayfs.  This hasnt been and isnt an issue, but a small cleanup I noticed could be made while testing builds on fargate for something else.


TerryHowe (approved):
/lgtm

PR #1568, opened 2022-12-19
Adds run-in-docker helps and better checksum + attribution generation

jaxesn:
*Issue #, if available:*

*Description of changes:*

The primary purpose of this PR is to better reuse the attribution and checksum generated files when running the "all" target which builds across release branches. With this change release branch'd components which are the same versions will not be rebuilt/generated instead the checksum and attribution intermediate files will be copied between release branch folders.

Also adds:
- New `run-<target>-in-docker` syntax which is variable driven making it easy to expose other targets via the docker container
- The docker container setup process now uses a bind mount instead of rsync which simplifies, removes, the need to pull files out explicitly since the host fs matches the containers.
- Small changes I made in eks-a to cleanup some of the output of create-binaries/cgo-create-binaries.
- `all-attributions` `all-checksums` `all-attributions-checksums` to make rebuilding for all release branches easier and these targets will reuse the intermediate files mentioned above.

Commits:
- Main changes
- Help.mk file regeneration adding the new targets to all the Help.mk files to support tab-completion
- Readme changes changing the attribution-checkum target name


Common.mk
```diff
@@ -459,9 +451,30 @@ define BUILDCTL
...
-			--output type=$(IMAGE_OUTPUT_TYPE),oci-mediatypes=true,\"name=$(ALL_IMAGE_TAGS)\",$(IMAGE_OUTPUT); \
-	fi
-endef 
+			--output type=$(IMAGE_OUTPUT_TYPE),oci-mediatypes=true,\"name=$(ALL_IMAGE_TAGS)\",$(IMAGE_OUTPUT)
+endef
+
+define CGO_DOCKER
+	source $(BUILD_LIB)/common.sh && build::docker::retry_pull --platform $(IMAGE_PLATFORMS) $(BUILDER_IMAGE); \
+	INTERACTIVE="$(shell if [ -t 0 ]; then echo '-it'; fi)"; \
+	docker run --rm $$INTERACTIVE -w /eks-anywhere-build-tooling/projects/$(COMPONENT) $(DOCKER_USE_ID_FOR_LINUX) \
```
markapruett, line 460: Should this be `eks-anywhere-build-tooling`?

build/lib/run_target_docker.sh
```diff
@@ -37,41 +37,15 @@ echo "****************************************************************"
...
-	--include="projects/$PROJECT/***" $EXTRA_INCLUDES \
-	--include='*/' --exclude='projects/***' $MAKE_ROOT/ eks-d-builder:/eks-distro
-
-# Need so git properly finds the root of the repo
-CURRENT_HEAD="$(cat $MAKE_ROOT/.git/HEAD | awk '{print $2}')"
-docker exec -it eks-d-builder mkdir -p /eks-distro/.git/{refs,objects} /eks-distro/.git/$(dirname $CURRENT_HEAD)
-docker cp $MAKE_ROOT/.git/HEAD eks-d-builder:/eks-distro/.git
-docker cp $MAKE_ROOT/.git/$CURRENT_HEAD eks-d-builder:/eks-distro/.git/$CURRENT_HEAD
-
-docker exec -it eks-d-builder make $TARGET -C /eks-distro/projects/$PROJECT RELEASE_BRANCH=$RELEASE_BRANCH IMAGE_REPO=$IMAGE_REPO ARTIFACTS_BUCKET=$ARTIFACTS_BUCKET
```
markapruett, line 77: Why can we get rid of all this?

build/update-attribution-files/make_attribution.sh
```diff
@@ -50,7 +49,32 @@ RELEASE_FOLDER=$(find $PROJECT_ROOT -type d -name "1-*")
...
+            # rebuilding/regenerating
+            if [[ $TARGET == *"checksums"* ]]; then
+                mkdir -p $PROJECT_ROOT/_output/$release
+                cp -rf $PROJECT_ROOT/_output/$LAST_RELEASE_BRANCH/bin $PROJECT_ROOT/_output/$release
+            fi
+            if [[ $TARGET == *"attribution"* ]]; then
+                mkdir -p $PROJECT_ROOT/_output/$release
+                cp -rf $PROJECT_ROOT/_output/$LAST_RELEASE_BRANCH/{attribution,LICENSES} $PROJECT_ROOT/_output/$release
+            fi
+            build::attribution::generate $release
```
markapruett, line 73: Do we need to run this again, if the git tag hasn't changed?

Common.mk
jaxesn: fixed

build/update-attribution-files/make_attribution.sh
jaxesn: We could do it another way, but the way this implements it, yes.  The intermediate files (LICESES/attribution/bin) are copied between the release branch output folders so the target still needs to run to generate the checksum file/copy attribution from _output to actual release_branch folder.  They ending up being super quick.

markapruett (approved):

markapruett (approved):

PR #1582, opened 2022-12-22
Bump iam-authenticator version

zafs23:
*Issue #, if available:*

*Description of changes:*


zafs23:
Do not need to upgrade the iam-auth version and make a patch. [Reference](https://gitlab.aws.dev/eks-dataplane/eks-kcp-ami-config/-/blob/mainline/configuration/global.yml#L128)

PR #1586, opened 2022-12-23
Bump etcd version to v3.5.6 for 1.22-1.24

zafs23:
*Issue #, if available:*

*Description of changes:*


projects/etcd-io/etcd/1-21/GIT_TAG
```diff
@@ -1 +1 @@
-v3.4.21
+v3.4.20
```
jaxesn, line 1: Let's not downgrade, in this case I know we wanted to upgrade to solve a bug in etcd, but even if we did it by accident I don't think we ever want to downgrade.

PR #1618, opened 2022-12-30
Make variable name consistent

markapruett:
*Issue #, if available:*

*Description of changes:*
The variable is called `ARTIFACTS_BUCKET` and `ARTIFACT_BUCKET` in different places. Add a check to ensure it's the same when used

### DIFF
--- a/Common.mk
+++ b/Common.mk
@@ -26,7 +26,8 @@ OUTPUT_BIN_DIR?=$(OUTPUT_DIR)/bin/$(REPO)
 #################### AWS ###########################
 AWS_REGION?=us-west-2
 AWS_ACCOUNT_ID?=$(shell aws sts get-caller-identity --query Account --output text)
-ARTIFACTS_BUCKET?=my-s3-bucket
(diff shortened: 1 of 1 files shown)


Common.mk
```diff
@@ -26,7 +26,7 @@ OUTPUT_BIN_DIR?=$(OUTPUT_DIR)/bin/$(REPO)
 #################### AWS ###########################
 AWS_REGION?=us-west-2
 AWS_ACCOUNT_ID?=$(shell aws sts get-caller-identity --query Account --output text)
-ARTIFACT_BUCKET?=my-s3-bucket # TODO: in eks-a we use ARTIFACTS_BUCKET
+ARTIFACTS_BUCKET?=my-s3-bucket
```
danbudris, line 29: is there any place, such as in the Prow jobs, where we're over-riding this variable using the `ARTIFACT_BUCKET` name?

markapruett: Yes -- loads. Good catch! I'll update there

markapruett: Updated to avoid changing in so many places at once

PR #1730, opened 2023-02-07
Have go mod & go sum generate correctly

markapruett:
*Issue #, if available:*

*Description of changes:*
Currently we don't copy over go.mod/sum if that version is the same because we never checked on the make target (update-go-mods) and it bails out since it doesn't match "checksums" or "attributions". 

Doing this will pull in the files correctly. 

### DIFF
--- a/projects/etcd-io/etcd/Makefile
+++ b/projects/etcd-io/etcd/Makefile
@@ -5,7 +5,7 @@ GOLANG_VERSION?=$(shell cat $(RELEASE_BRANCH)/GOLANG_VERSION)
 REPO=etcd
 REPO_OWNER=etcd-io
 
-BINARY_TARGET_FILES=server etcdctl
+BINARY_TARGET_FILES=etcd etcdctl
 SOURCE_PATTERNS=go.etcd.io/etcd ./etcdctl
 EXTRA_GOBUILD_FLAGS=-installsuffix cgo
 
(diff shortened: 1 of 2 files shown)


projects/etcd-io/etcd/Makefile
```diff
@@ -5,7 +5,7 @@ GOLANG_VERSION?=$(shell cat $(RELEASE_BRANCH)/GOLANG_VERSION)
 REPO=etcd
 REPO_OWNER=etcd-io
 
-BINARY_TARGET_FILES=etcd etcdctl
+BINARY_TARGET_FILES=server etcdctl
```
jaxesn, line 8: This should be defined https://github.com/aws/eks-distro/blob/main/projects/etcd-io/etcd/1-22/Overrides.mk#L5

When 1.22 was introduced we were building older versions which needed `etcd etcdctl` to be the value here so the Makefile was kept as is for those versions and 1.22 was the special case with the Overrides.mk.  If we aren't building the version anymore that 1.21 used to have, we should change the Makefile to default to the 1.22 setup and remove the overrides.  If we havent included the overrides.mk in newer versions we should include it.  If we have and the value from the overrides is not overwriting this, we should track down why

jaxesn (approved):

markapruett:
Holding until after release today

PR #1732, opened 2023-02-07
support for building EKS-D in codebuild

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-distro-internal/issues/269

*Description of changes:*
Changes required to support EKS-D builds in Codebuild.

One big change is splitting up the `kOps` targets in the makefile to support both `kOps` invoked from Prow and `kOps` invoked from Codebuild. Due to differences in the environments, and the way they handle credentials, we have to use a slightly different entrypoint script for Codebuild than Prow. Namely, codebuild jobs are running in ECS containers with a given role (in this case, the codebuild standard role we set up for the pipeline), rather than a pod with IRSA, so we have to set up the local AWS profile accordingly (https://github.com/aws/eks-distro/pull/1732/files#diff-85d9cebfc642bcd6756fd827404cdfcc79e237e3c3681ff8b7d1140dcb74481aR32). There are also changes to the paths we're outputing to in the common.mk that differ from the way that EKS-A is using codebuild; going to try and reconcile this for ease of porting over those common scripts.


Common.mk
```diff
@@ -49,12 +49,11 @@ CODEBUILD_BUILD_IMAGE?=
 CLONE_URL?=$(call GET_CLONE_URL,$(REPO_OWNER),$(REPO))
 #HELM_CLONE_URL=$(call GET_CLONE_URL,$(HELM_SOURCE_OWNER),$(HELM_SOURCE_REPOSITORY))
 HELM_CLONE_URL=https://github.com/$(HELM_SOURCE_OWNER)/$(HELM_SOURCE_REPOSITORY).git
+ARTIFACTS_PATH?=$(MAKE_ROOT)/_output/tar
```
jaxesn, line 52: If we can avoid changing this file at all that would be best, since we try to sync this between eks-a as well.

Could we just set the ARTIFACTS_PATH in our code build spec?

danbudris: probably, i'll take a look

build/lib/common.sh
```diff
@@ -278,13 +278,8 @@ function build::common::get_clone_url() {
...
   local -r repo=$2
   local -r aws_region=$3
-  local -r codebuild_ci=$4
 
-  if [ "$codebuild_ci" = "true" ]; then
-    echo "https://git-codecommit.${aws_region}.amazonaws.com/v1/repos/${org}.${repo}"
-  else
-    echo "https://github.com/${org}/${repo}.git"
-  fi
+  echo "https://github.com/${org}/${repo}.git"
```
markapruett, line 282: Is this because we don't keep anything in codecommit anymore?

danbudris: yeah -- we don't. EKS-A does, and this script is ported over from EKS-A

Makefile
```diff
@@ -69,36 +69,45 @@ postsubmit-build: setup
...
+kops-amd: export UBUNTU_RELEASE=focal-20.04
+kops-amd: export UBUNTU_RELEASE_DATE=20221018
+kops-amd: kops-prereqs
+	RELEASE=$(RELEASE) $(KOPS_ENTRYPOINT)
+
+.PHONY: kops-arm
+kops-arm: export NODE_INSTANCE_TYPE=t4g.medium
+kops-arm: export NODE_ARCHITECTURE=arm64
+kops-arm: export UBUNTU_RELEASE=focal-20.04
+kops-arm: export UBUNTU_RELEASE_DATE=20221018
```
kschumy, line 96: It seems like we have some repeated exported values (like `focal-20.04` and `20221018`). Are these always going to be the same for amd and arm. If so, should be use variables we set at the top?

danbudris: ok, de-duped it by setting the var at the top level of the `kops` target, so it'll default to those releases, and then we can over-ride it in sub-targets as needed

markapruett (approved):

Makefile
```diff
@@ -69,36 +69,43 @@ postsubmit-build: setup
...
+kops: $(if $(CODEBUILD_BUILD_ID),kops-codebuild,kops-prow)
+
+.PHONY: kops-codebuild
+kops-codebuild: KOPS_ENTRYPOINT=development/kops/codebuild.sh
+kops-codebuild: kops-amd kops-arm
+	@echo 'Done kops-codebuild'
+
+.PHONY: kops-prow
+kops-prow: KOPS_ENTRYPOINT=development/kops/prow.sh
+kops-prow: kops-amd kops-arm kops-arm-ubuntu-22
```
markapruett, line 85:
```suggestion
kops-prow: kops-amd kops-arm
```

PR #1751, opened 2023-02-09
Improve build log out but still allow full debug log

jaxesn:
*Issue #, if available:*

*Description of changes:*

This builds on something weve in [build-tooling](https://github.com/aws/eks-distro-build-tooling/blob/main/eks-distro-base/scripts/eks-d-common#L53) to limit the output when building the minimal images, but still allow an override via the `OUTPUT_DEBUG_LOG` env var to see it all if you want.

This PR introduces `build::common::echo_and_run` along with start and end log message for key targets in Common.mk.  echo_and_run was added to a few key places in our common scripts and the `set -x` were removed.  The resulting output, imo, is easier to read and follow and should still provide key data points for when scripts fail. That said, I do expect to see failures in the future that we want to add more logging for. In those case, we can temporary set the `OUTPUT_DEBUG_LOG` env var to see the full debug log.

IMO, set -x is nice because it gives you all the output, but if you are not intimately familiar with and have a strong understanding of what these scripts are doing, its too overwhelming and introduces tons of red herrings.  

Please review the output from some of the presubmit jobs to see what you think and lets discuss!

### DIFF
--- a/Common.mk
+++ b/Common.mk
@@ -221,7 +221,7 @@ trimap = $(and $(strip $2),$(strip $3),$(strip $4),$(call \
 
 _pos = $(if $(filter $1,$2),$(call _pos,$1,\
        $(call list-rem,$2),x $3),$3)
-pos = $(words $(call _pos,$1,$2))
+pos = $(words $(call _pos,$1,$2,))
 
 # TODO: this exist in the gmsl, https://gmsl.sourceforge.io/
 # look into introducting gmsl for things like this
@@ -425,6 +425,13 @@ SKIP_CHECKSUM_VALIDATION?=false
 IN_DOCKER_TARGETS=all-attributions all-attributions-checksums all-checksums attribution attribution-checksums binaries checksums clean clean-go-cache
 ####################################################
 
+#################### LOGGING #######################
+DATE_CMD=TZ=utc $(shell if [ "$$(uname -s)" = "Darwin" ] && command -v gdate &> /dev/null; then echo gdate; else echo date; fi)
+DATE_NANO=$(shell if [ "$$(uname -s)" = "Linux" ] || command -v gdate &> /dev/null; then echo %3N; fi)
+TARGET_START_LOG?=$(eval _START_TIME:=$(shell $(DATE_CMD) +%s.$(DATE_NANO)))\\n------------------- $(shell $(DATE_CMD) +"%Y-%m-%dT%H:%M:%S.$(DATE_NANO)%z") Starting $@ target ------------------- 
+TARGET_END_LOG?="------------------- `$(DATE_CMD) +'%Y-%m-%dT%H:%M:%S.$(DATE_NANO)%z'` Finished $@ target (duration `echo $$($(DATE_CMD) +%s.$(DATE_NANO)) - $(_START_TIME) | bc` seconds) -------------------\\n"
+####################################################
+
 #################### TARGETS FOR OVERRIDING ########
 BUILD_TARGETS?=validate-checksums attribution $(if $(IMAGE_NAMES),local-images,) $(if $(filter true,$(HAS_HELM_CHART)),helm/build,) $(if $(filter true,$(HAS_S3_ARTIFACTS)),upload-artifacts,) attribution-pr
 RELEASE_TARGETS?=validate-checksums $(if $(IMAGE_NAMES),images,) $(if $(filter true,$(HAS_HELM_CHART)),helm/push,) $(if $(filter true,$(HAS_S3_ARTIFACTS)),upload-artifacts,)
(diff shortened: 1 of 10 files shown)


Common.mk
```diff
@@ -221,7 +221,7 @@ trimap = $(and $(strip $2),$(strip $3),$(strip $4),$(call \
 
 _pos = $(if $(filter $1,$2),$(call _pos,$1,\
        $(call list-rem,$2),x $3),$3)
-pos = $(words $(call _pos,$1,$2))
+pos = $(words $(call _pos,$1,$2,))
```
jaxesn, line 224: this was just a long standing bug that only shows it self in a couple builds, and even then its just a warning but its confusing so im fixing it.

danbudris:
I really like this, and I like the idea of having the ability to set `-x` for debugging purposes; we can then add more logging as needed based on waht we find with the debugging enabled.

Common.mk
```diff
@@ -425,6 +425,11 @@ SKIP_CHECKSUM_VALIDATION?=false
 IN_DOCKER_TARGETS=all-attributions all-attributions-checksums all-checksums attribution attribution-checksums binaries checksums clean clean-go-cache
 ####################################################
 
+#################### LOGGING #######################
+TARGET_START_LOG?=$(eval _START_TIME:=$(shell date +%s))\\n------------------- $(shell date +"%Y-%m-%dT%H:%M:%S%z") Starting $@ target -------------------
```
danbudris, line 429: the `$@`, that's an automatic variable which resolves to the name of the target this is executed in, is that right?

danbudris, line 429: should we set `TZ=utc` in here?

danbudris, line 429: some of these targets are super quick, add tenths of a second?

```suggestion
TARGET_START_LOG?=$(eval _START_TIME:=$(shell date +%s))\\n------------------- $(shell date +"%Y-%m-%dT%H:%M:%S:%1N%z") Starting $@ target ------------------- 
```

Common.mk
```diff
@@ -425,6 +425,11 @@ SKIP_CHECKSUM_VALIDATION?=false
 IN_DOCKER_TARGETS=all-attributions all-attributions-checksums all-checksums attribution attribution-checksums binaries checksums clean clean-go-cache
 ####################################################
 
+#################### LOGGING #######################
+TARGET_START_LOG?=$(eval _START_TIME:=$(shell date +%s))\\n------------------- $(shell date +"%Y-%m-%dT%H:%M:%S%z") Starting $@ target ------------------- 
+TARGET_END_LOG?="------------------- `date +'%Y-%m-%dT%H:%M:%S%z'` Finished $@ target (duration `echo $$[( $$(date +%s)-$(_START_TIME) )]` seconds) -------------------\\n"
```
danbudris, line 430:
```suggestion
TARGET_END_LOG?="------------------- `date +'%Y-%m-%dT%H:%M:%S:%1N%z'` Finished $@ target (duration `echo $$[( $$(date +%s)-$(_START_TIME) )]` seconds) -------------------\\n"
```

Common.mk
jaxesn: yep

jaxesn: will do

build/lib/buildkit_check.sh
```diff
@@ -27,5 +27,6 @@ function buildkit_ready() {
   echo "Buildkit daemon is not available"
   exit 1
 }
+yum install bc -y
```
jaxesn, line 30: i will move this to the builder-base build

Common.mk
```diff
@@ -425,6 +425,13 @@ SKIP_CHECKSUM_VALIDATION?=false
 IN_DOCKER_TARGETS=all-attributions all-attributions-checksums all-checksums attribution attribution-checksums binaries checksums clean clean-go-cache
 ####################################################
 
+#################### LOGGING #######################
+DATE_CMD=TZ=utc $(shell if [ "$$(uname -s)" = "Darwin" ] && command -v gdate &> /dev/null; then echo gdate; else echo date; fi)
+DATE_NANO=$(shell if [ "$$(uname -s)" = "Linux" ] || command -v gdate &> /dev/null; then echo %N; fi)
```
danbudris, line 430: is three decimal places enough?

```suggestion
DATE_NANO=$(shell if [ "$$(uname -s)" = "Linux" ] || command -v gdate &> /dev/null; then echo %3N; fi)
```

Common.mk
danbudris: hahah that sure is something

Common.mk
jaxesn: done

danbudris (approved):

markapruett (approved):

PR #1904, opened 2023-03-30
Revert "Added make command to help with minor release docs generation"

zafs23:
Reverts aws/eks-distro#1901


kschumy:
Why are we reverting this? If there are problems with it, is there a reason why we cannot fix them? Is the effort too much to fix them in comparison to the benefit we get from this?

PR #1926, opened 2023-04-13
Updated to Kubernetes 1.27.0

kschumy:
*Issue #, if available:*

*Description of changes:*


kschumy:
fixed merge conflicts

kschumy:
/hold 

we might not be releasing this patch version

projects/kubernetes/kubernetes/build/lib/binaries.sh
```diff
@@ -21,33 +21,39 @@ function build::binaries::kube_bins() {
...
-        cmd/kube-controller-manager \
-        cmd/kube-scheduler"
+  cd $repository
+  
+  
+  # kube build scripts set GOCACHE and GOMODCACHE to be in the ./_output/local/go/cache directory
+  # relative to the kubernetes source folder instead
+  # unsetting these two which we set in our common script to allow kube to fully manage these locations
+  # and keep the directories we need to clean consistent through versions
+  unset GOCACHE GOMODCACHE
```
jaxesn, line 31: prior to 1.27 they always set GOCACHE in the kube bin script. In 1.27 they allow the user to override this, which we do by default in our scripts. This was making our cleanup below not complete. This feels like the easiest fix and the one that is most consistent with how we have used these kube scripts.

kschumy:
We're moving forward with #1956

PR #1934, opened 2023-04-13
Fix out-of-space for build

kschumy:
*Issue #, if available:*

*Description of changes:*
* Fix out-of-space build issue from https://github.com/aws/eks-distro/pull/1926
* Intellij auto fixed whitespaces


projects/kubernetes/kubernetes/build/lib/binaries.sh
```diff
@@ -63,14 +63,18 @@ function build::binaries::kube_bins() {
 	hack/make-rules/build.sh -trimpath cmd/kubelet \
```
jaxesn, line 63: can u tab this line in one so its more clear that its separate from the export

jaxesn (approved):

PR #2002, opened 2023-05-01
Adds ecr credential provider image

markapruett:
*Issue #, if available:*

*Description of changes:*
Adds ecr credential provider build from aws-ccm


projects/kubernetes/cloud-provider-aws/Makefile
```diff
@@ -5,21 +5,23 @@ GOLANG_VERSION?=$(shell cat ./$(RELEASE_BRANCH)/GOLANG_VERSION)
...
+BINARY_TARGET_FILES=aws-cloud-controller-manager ecr-credential-provider
+SOURCE_PATTERNS=./cmd/aws-cloud-controller-manager ./cmd/ecr-credential-provider
 
 KUBE_BASE_GIT_TAG?=$(shell cat $(BASE_DIRECTORY)/projects/kubernetes/release/$(RELEASE_BRANCH)/GIT_TAG)
 KUBE_BASE_TAG?=${KUBE_BASE_GIT_TAG}-eks-${RELEASE_BRANCH}-${RELEASE}
 GO_RUNNER_IMAGE?=$(IMAGE_REPO)/kubernetes/go-runner:$(KUBE_BASE_TAG)
 BASE_IMAGE=$(GO_RUNNER_IMAGE)
 
 HAS_RELEASE_BRANCHES=true
+HAS_S3_ARTIFACTS=true
```
markapruett, line 17: This will produce tarballs and images for both. Not sure if we have this case solved elsewhere, but we don't need the binaries for aws-cloud-controller-manager. Either way, doesn't seem like a big deal.

PR #2003, opened 2023-05-02
Updated go version for Kunbernetes 1.23 and 1.24

kschumy:
*Issue #, if available:*

*Description of changes:*
* 1.23: https://github.com/kubernetes/kubernetes/blob/v1.23.17/build/build-image/cross/VERSION
* 1.24: https://github.com/kubernetes/kubernetes/blob/v1.24.13/build/build-image/cross/VERSION

As wild as this sounds, the checksums for 1.24 do not change when changing the go version to 1.19. Because I simply did not believe it, I rolled back the 1.24 patch version by one and had it build with 1.19. Then I compared the generated checksums with the [checksums](https://github.com/aws/eks-distro/blame/93d6882ce4b2e91234ade5378dd064a7dbc7cb48/projects/kubernetes/kubernetes/1-24/CHECKSUMS) we had for that version when we build it with 1.18. They matched. I still find it hard to believe, but apparently the binaries are EXACTLY the same when building with go 1.18 and 1.19.

---


projects/kubernetes/kubernetes/1-23/GOLANG_VERSION
```diff
@@ -1 +1 @@
-1.17
+1.19
```
markapruett, line 1: 🥳

markapruett (approved):

PR #2007, opened 2023-05-02
Bump golang/x/net to 0.7.0 and golang/x/text to v0.3.8 for CoreDNS 1.22-1.27

zafs23:
*Issue #, if available:*

*Description of changes:*Fixes [CVE 41723](https://github.com/advisories/GHSA-vvpx-j8f3-3w6h). Any `golang/x/net `dependency < 0.7.0 is vulnerable to this CVE. 
Fixes [CVE-2022-32149](https://nvd.nist.gov/vuln/detail/cve-2022-32149) by bumping golang/x/text to v0.3.8
These patches are temporary fix until EKS update to the latest release where these CVEs are fixed. 

### DIFF
--- a/projects/coredns/coredns/1-22/patches/0001-Bump-golang-x-net-and-golang-x-text-dependency.patch
+++ b/projects/coredns/coredns/1-22/patches/0001-Bump-golang-x-net-and-golang-x-text-dependency.patch
@@ -1,0 +1,201 @@
+From 70561bf681e8e21f48b9ebfcf7daca8cc3ef7e95 Mon Sep 17 00:00:00 2001
+From: Sajia Zafreen <szafreen@amazon.com>
+Date: Tue, 2 May 2023 21:56:39 +0000
+Subject: [PATCH] Bump golang/x/net and golang/x/text dependency
+
+Signed-off-by: Sajia Zafreen <szafreen@amazon.com>
+---
+ go.mod |  7 ++++-
+ go.sum | 86 +++++++++++-----------------------------------------------
+ 2 files changed, 22 insertions(+), 71 deletions(-)
+
+diff --git a/go.mod b/go.mod
+index a72ab760..d401e1df 100644
+--- a/go.mod
++++ b/go.mod
+@@ -2,6 +2,11 @@ module github.com/coredns/coredns
+ 
+ go 1.16
+ 
++replace (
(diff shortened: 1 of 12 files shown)


projects/coredns/coredns/1-22/patches/0001-Bump-golang-x-net-and-golang-x-text-dependency.patch
```diff
@@ -0,0 +1,298 @@
...
+
+diff --git a/go.mod b/go.mod
+index a72ab760..38db0091 100644
+--- a/go.mod
++++ b/go.mod
+@@ -1,13 +1,16 @@
+ module github.com/coredns/coredns
+ 
+-go 1.16
++go 1.17
```
markapruett, line 20: I don't think this will cause issues because newer versions are backwards compatible.

projects/coredns/coredns/1-22/patches/0001-Bump-golang-x-net-and-golang-x-text-dependency.patch
```diff
@@ -0,0 +1,298 @@
...
++	github.com/spf13/pflag v1.0.5 // indirect
++	github.com/stretchr/objx v0.2.0 // indirect
++	github.com/tinylib/msgp v1.1.2 // indirect
++	go.etcd.io/etcd/client/pkg/v3 v3.5.1 // indirect
++	go.opencensus.io v0.23.0 // indirect
++	go.uber.org/atomic v1.7.0 // indirect
++	go.uber.org/multierr v1.6.0 // indirect
++	go.uber.org/zap v1.17.0 // indirect
++	golang.org/x/mod v0.6.0-dev.0.20220419223038-86c51ed26bb4 // indirect
++	golang.org/x/net v0.0.0-20220722155237-a158d28d115b // indirect
```
markapruett, line 115: Not sure this will work according to https://go.dev/ref/mod#go-mod-file-replace. 

> Note that a replace directive alone does not add a module to the [module graph](https://go.dev/ref/mod#glos-module-graph). A [require directive](https://go.dev/ref/mod#go-mod-file-require) that refers to a replaced module version is also needed, either in the main module’s go.mod file or a dependency’s go.mod file. A replace directive has no effect if the module version on the left side is not required.

markapruett (approved):

PR #2024, opened 2023-05-10
Update cni plugins

kschumy:
*Issue #, if available:*

*Description of changes:*
~~For some reason, go mod tidy is generating an error. Will work on resolving this~~


kschumy:
Locally, had run into this exact issue https://github.com/safchain/ethtool/issues/57

kschumy:
testing two different solutions out on 1.22 and 1.23. I think 1.23  is correct, but I want to make sure

kschumy:
The 1.23 approach, which also updated vendor, was correct. The failure was due to checksums. Updating all now...

projects/containernetworking/plugins/1-22/patches/0001-Bump-golang-x-net-to-0.7.0.patch
zafs23, line 1: Need to bump golang/x/text to 0.3.8 in the patch to fix CVE-2021-38561

zafs23: @kschumy does the patch need to have go 1.18 instead of go1.17?

kschumy: So we should be good not handling that other CVE, right?

zafs23: @kschumy it's not the CVE, the golang version is generated with 1.18 but in the go.mod it is 1.17

kschumy: we have that for several projects

Issue #2250, opened 2023-08-04
Remove deprecated flag `--short` from `kubectl version` calls in `kOps` execution against `v1.28` and elsewhere


danbudris:
the `kubectl version` flag `--short` has been deprecated. The default output of `kubectl version` has changed to be that of `--short`. This is showing up as an error in the `v1.28` build logs during the `kOps` builds. 

This is in at least `install_requierments.sh`, but should be ferreted out elsewhere.

Remove the `kubectl veresion` `--short`, cased for release branch `v1.28`.

PR #2289, opened 2023-08-17
remove deprecated flag --short

MjMoshiri:
*Issue #, if available:*
#2250 
*Description of changes:*

As indicated in the Kubernetes [changelog](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.28.md#deprecation), the --short flag has been eliminated, and the default output will now match what was previously displayed with the short option.

This PR removes the utilization of this flag from the source code.


development/kops/install_requirements.sh
```diff
@@ -69,7 +69,7 @@ then
 fi
 
 echo "$(which kops): $(kops version)"
-echo "$(which kubectl): $(kubectl version --client=true --short)"
+echo "$(which kubectl): $(kubectl version --client=true)"
```
rcrozean, line 72: This should check which version `$(RELEASE_BRANCH)` is being tested. Only `1-28` deprecates `--short`

MjMoshiri: like this?

```
if [[ "${RELEASE_BRANCH}" == "1-28" ]]; then
    echo "$(which kubectl): $(kubectl version --client=true)"
else
    echo "$(which kubectl): $(kubectl version --client=true --short)"
fi
```

if yes, what would happen with the next versions?

rcrozean: The deprecation occurred with 1.28+ I would recommend something like this. It will use the `--short` flag on all versions that still use it without needing to update with future versions.
```
if [[ $test == "1-23" || $test == "1-24" || $test == "1-25" || $test == "1-26" || $test == "1-27" ]]; then
  echo "kubectl version --client=true --short"
else
  echo "kubectl version --client=true"
fi
```

MjMoshiri:
@rcrozean Hi, 
I have added the Would you review? 🙏🏻

rcrozean:
Thanks for the update. We are adding some tests to this portion of the repo.

PR #2407, opened 2023-09-26
Bumped development release number for 1.28

rcrozean:
*Issue #, if available:*

*Description of changes:*


bhavi-koduru:
This commit message is for the initial release when we add a new file, wondering why it is using the same for next releases as well. Do this get updated after official release?

PR #2421, opened 2023-10-03
Add commands to test development/kops/install_requirements.sh

rcrozean:
*Issue #, if available:*

*Description of changes:*
Add make targets for testing the files in `development/kops/*.sh` like [development/kops/install_requirements.sh](https://github.com/aws/eks-distro/blob/main/development/kops/install_requirements.sh). These files are important to our build tasks and should be tested validity when changed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ kops-arm-ubuntu-22: kops-prereqs
 	RELEASE=$(RELEASE) $(KOPS_ENTRYPOINT);
 
 .PHONY: kops-prereqs
-kops-prereqs: $(if $(JOB_TYPE) == "presubmit",,postsubmit-build)
+kops-prereqs: $(if $(filter presubmit,$(JOB_TYPE)),,postsubmit-build)
 	ssh-keygen -b 2048 -t rsa -f ~/.ssh/id_rsa -q -N ""
 	cd development/kops && RELEASE=$(RELEASE) ./install_requirements.sh
 
--- a/Makefile
+++ b/Makefile
@@ -104,35 +104,7 @@ kops-arm-ubuntu-22: kops-prereqs
 	RELEASE=$(RELEASE) $(KOPS_ENTRYPOINT);
 
 .PHONY: kops-prereqs
-kops-prereqs: postsubmit-build
-	ssh-keygen -b 2048 -t rsa -f ~/.ssh/id_rsa -q -N ""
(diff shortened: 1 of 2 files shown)


development/kops/run_all.sh
```diff
@@ -17,44 +17,46 @@ set -exo pipefail
...
-      createDate=$(date -d"$createDate" +%s)
-      olderThan=$(date --date "1 day ago" +%s)
-      if [[ $createDate -lt $olderThan ]]; then
-        echo "Deleting cluster ${cluster_fqdn}"
-        ${KOPS} delete cluster --state "${KOPS_STATE_STORE}" --name ${cluster_fqdn} --yes
-        aws s3 rm --recursive "${KOPS_STATE_STORE}/${cluster_name}"
-      fi
-    fi
-  done
+if [[ "${KOPS_STATE_STORE}" != "" && "$JOB_TYPE" != "presubmit" ]]; then
```
jaxesn, line 36: i might would reorder this file a bit if you can to avoid havng to check this twice.

Makefile
```diff
@@ -104,7 +104,7 @@ kops-arm-ubuntu-22: kops-prereqs
 	RELEASE=$(RELEASE) $(KOPS_ENTRYPOINT);
 
 .PHONY: kops-prereqs
-kops-prereqs: postsubmit-build
+kops-prereqs: $(if $(JOB_TYPE) == "presubmit",,postsubmit-build)
```
jaxesn, line 107: this is not quite right...  you probably want this

`$(if $(filter presubmit,$(JOB_TYPE)),,postsubmit-build)`

Check Common.mk youll see this kind of thing.

jaxesn (changes requested):

Makefile
rcrozean: Used [open_pr_check.sh](https://github.com/aws/eks-distro-build-tooling/blob/64e9741d152acddf6b6aa54cdbae77ac728232be/pr-scripts/open_pr_check.sh#L26) as the example. Will update to your recommendation. Don't see us doing this in eks-distro, other than a single call in the [buildkit.sh](https://github.com/aws/eks-distro-build-tooling/blob/64e9741d152acddf6b6aa54cdbae77ac728232be/scripts/buildkit.sh#L92) where we trim, but not sure why this is the only example of us trimming in that repo.

jaxesn (approved):

PR #3089, opened 2024-06-04
add docker tars to server.tgz for k/k

jaxesn:
*Issue #, if available:*

*Description of changes:*

While looking at using the new kind build-image support Dims added: https://github.com/kubernetes-sigs/kind/commit/e29f332cbb2412d683e357b4cff5cdbe9b906b16 I noticed there are a couple differences from the eks-d server.tgz and the upstream:

- the 4 server component images are included next to the bins themselves
- these image tars are in [docker](https://github.com/kubernetes/kubernetes/blob/ae5543e4c8f99cb1555102a8ebc310aed3c82596/build/lib/release.sh#L364) format instead of oci
- kubectl is included

source.tgz is also included, but that is not needed for the kind flow i was testing so i have not added that to the server.tgz.

Per the buildkit docs the docker format is still [compat](https://github.com/moby/buildkit?tab=readme-ov-file#docker-tarball) with oci so in the cases where the existing image was being used, for example in the eks-a image build process, they should still import just fine. I verified this using the kind build, which imports it the same way using ctr.

The final server.tgz will be bigger, but I think this is a reasonable change to match upstream and make downstream usage of kind in eks-a simpler and more standard.


projects/kubernetes/kubernetes/Makefile
```diff
@@ -100,6 +110,7 @@ cleanup-artifacts:
 	rm -rf $(OUTPUT_DIR)/{attribution,LICENSES,pause} $(OUTPUT_DIR)/bin/{LICENSES,ATTRIBUTION.txt}
 	mv $(OUTPUT_DIR)/bin/* $(ARTIFACTS_PATH)/bin
 	cp $(OUTPUT_DIR)/ATTRIBUTION.txt $(ARTIFACTS_PATH)
+	if [ "$(FAKE_ARM_IMAGES_FOR_VALIDATION)" = "true" ]; then rm -rf $(OUTPUT_DIR)/images/bin/linux/arm64; fi
```
bhavi-koduru, line 113: is `FAKE_ARM_IMAGES_FOR_VALIDATION` meant to be set somewhere?

bhavi-koduru: i see it here https://github.com/aws/eks-distro/blob/2f71476c12d964cdd2f249cd3ceb76442b93aee0/Common.mk#L414

PR #3119, opened 2024-06-12
Kubernetes Project Updates

vineeth-bandi:
*Issue #, if available:*

*Description of changes:*
Kubernetes project updates for 1.27 - 1.30


projects/kubernetes/kubernetes/1-30/patches/0002-EKS-PATCH-Update-log-verbosity-for-node-health-and-t.patch
```diff
@@ -5,6 +5,7 @@ Subject: [PATCH] --EKS-PATCH-- Update log verbosity for node health and taint
  checks
 
 cherrypick of https://github.com/kubernetes/kubernetes/pull/121967
+For more details https://t.corp.amazon.com/V1088905731
```
rcrozean, line 8: We should remove this link

vineeth-bandi: Yeah that makes sense. I'll make that change for all of those updated patches.

PR #4827, opened 2026-07-25
fix: Raise kubernetes/release Go pins to 1.25

charles-large:
Build tooling commit 29732a46 removed every Go version below 1.25 from builder-base, so the toolchain directory that a 1.24 pin resolves to no longer exists in the image. The fleet first ran on such an image when eks-distro-prow-jobs commit da2bf1fa moved the pinned builder-base tag to standard-ec39a22a692a667411d582405edec7f191486e52.2, which is the first pinned tag built without Go 1.24.


varma-maryala:
lgtm

